PR #11, opened 2020-03-17
adding ability to add sidebar elements to left TOC

choldgraf:
This is a very hacky PR :-)

It adds the ability to control a few sidebar elements by using this syntax in a `{toctree}` directive:

````
```{toctree}
page1
page2
++ divider
++ header: My header
page3
```
````

This is a feature many people utilized in Jupyter Book, and I can't figure out how to have the same control over the sidebar without doing this.

It does this by modifying the TOCTREE source in-place and adding `http://` to it, which Sphinx will not try to process and instead just passes through at rendering. Then, there is some rendering logic in the sidebar that results in the proper HTML elements.

It is quite hacky, but also a nice feature. I am curious if folks think it is *too* hacky. @chrisjsewell or @rossbar, any thoughts?

**Chris edit: yeah this is definitely too hacky, it'll mess up the relative ordering of pages. We need to find a way to accomplish the same thing but actually removing the special items from the TOC tree**.

here's a link to the page that describes this in the docs: https://14-244496643-gh.circle-artifacts.com/0/html/layout.html#adding-special-items-to-the-toc

### CODE REVIEW

**Comment by chrisjsewell** [comment]
Maybe try patching the actual toctree directive:

```python
from docutils.parsers.rst import directives
from sphinx.directives.other import TocTree

class NewTocTree(TocTree):
   def parse_content(self, toctree: addnodes.toctree) -> List[Node]:
       self.content = self.content # do wht you need to do
       super().parse_content(toctree)

def setup(app: "Sphinx") -> Dict[str, Any]:
    directives.register_directive('toctree', NewTocTree)
```

choldgraf:
oof, I have put way more time than I should have into this haha, I don't know whether this is going to be possible utilizing sphinx pieces under the hood.

I've tried patching the `TocTree` directive in order to parse its contents for the `++ ` syntax first, but I'm not sure how to propagate that information to the HTML jinja step and also scrub the TocTree object of this information as well. I tried keeping track of the HTML page that comes next, but this wasn't enough information to reliably put these pieces in the sidebar. I feel like this is a rabbit hole that is gonna distract from more important things so I'm gonna put this on the backburner for now. I'll close this PR but open an issue to keep track of the item

PR #16, opened 2020-03-19
adding extra TOC item functionality

choldgraf:
This implements the ability to control some extra things in the left TOC using options in the `toctree` directive. It is slightly less-hacky than #11 but still a bit hacky :-)

Basically, we override the base TocTree directive, check it for a few extra options that we support, and then store them for use later when rendering the HTML. We also store the page on which the TOC was listed, as well as the first item of the TOCtree.

Then, when the sidebar HTML is rendered, we check each child and see if it was stored when a toctree special option was called. Here's the behavior that we currently support:

* If `:expand_sections:` is given in a toctree directive, then the children of the *parent page* will always be shown in the left TOC
* If `:divider:` is given, then a horizontal rule is placed before the *first item* of the toctree directive.
* If `:header:` is given, then a header is placed before the *first item* of the toctree directive.

I'm curious what @chrisjsewell thinks about this one...it's a little bit tricky but I think a bit cleaner than having people put things in the `toctree` content itself.

### DIFF
--- a/sphinx_book_theme/__init__.py
+++ b/sphinx_book_theme/__init__.py
@@ -2,11 +2,18 @@
 from pathlib import Path
 import docutils
 from myst_nb.parser import CellNode
+from docutils.parsers.rst import directives
 from sphinx.util import logging
+from sphinx.directives.other import TocTree
 import sass
 
 __version__ = "0.0.1dev0"
 SPHINX_LOGGER = logging.getLogger(__name__)
+EXTRA_TOC_OPTIONS = {
+    "expand_sections": directives.flag,
+    "caption": directives.unchanged_required,
+    "divider": directives.flag,
+}
 
 
 def get_html_theme_path():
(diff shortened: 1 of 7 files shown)


choldgraf:
also, see the built docs for how this looks:

https://30-244496643-gh.circle-artifacts.com/0/html/index.html

for reference, here are the toctrees that are defined on the first index page that let us do this:

````
```{toctree}
:maxdepth: 1
:header: Main docs
layout
notebooks
```
```{toctree}
:divider:
:header: Reference items
demochapter1/index
limits
```

````

File: sphinx_book_theme/__init__.py [hunk 1]
```diff
@@ -22,6 +29,28 @@ def add_static_path(app):
     app.config.html_static_path.append(str(static_path))
 
 
+def find_url_relative_to_root(pagename, relative_page, path_docs_source):
```
choldgraf: this is because when the sidebar is rendered, we only have the path to each item *relative to the current page*. whereas when we store page names in the toctree directive, they are always relative to the *site root*.

PR #24, opened 2020-03-29
fixing config key error

choldgraf:
This fixes a bug that @najuzilu reported

### DIFF
--- a/docs/notebooks.ipynb
+++ b/docs/notebooks.ipynb
@@ -304,7 +304,11 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "tags": [
+     "raises-exception"
+    ]
+   },
    "outputs": [],
    "source": [
     "this_will_error"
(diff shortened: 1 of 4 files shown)


najuzilu:
I figured out what is happening. The error occurs when I do not add `path_to_book : ""` under `repository`. Is this by design?

choldgraf:
It should be fixed by this PR... Does it still not work if you don't provide that config value at all?

najuzilu:
The example below throws this ValueError: `You must provide the key: repository_url to add Binder buttons.`
```
repository:
  url                       : ""
  path_to_book              : ""
```

This example throws the error I mentioned previously: `AttributeError: 'NoneType' object has no attribute 'strip'`
```
repository:
  url                       : https://github.com/ExecutableBookProject/quantecon-example/tree/master/book/docs
```

Whereas this example does not throw any errors:
```
repository:
  url                       : https://github.com/ExecutableBookProject/quantecon-example/tree/master/book/docs
  path_to_book              : ""
```

So, it looks like we're required to have both but we can't have one without the other. I was wondering if that's on purpose.

choldgraf:
ok give it another shot, I think it's fixed in #27

najuzilu:
It's working great now! Thanks @choldgraf

PR #51, opened 2020-04-14
figures and captions

choldgraf:
Updates CSS for figure captions, also adds rules for sidebar captions:

https://147-244496643-gh.circle-artifacts.com/0/html/reference/limits.html#common-publishing-items

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,6 +21,7 @@ templates_path = ["_templates"]
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", "**.ipynb_checkpoints"]
 
+numfig = True
 
 # -- Options for HTML output -------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


choldgraf:
@najuzilu I've sent an invite to add you as a maintainer on this repository, in case you'd like to give this a merge when you're happy with it. Otherwise I'll merge a bit later unless you have comments

najuzilu:
Nicely done! I think aligning figures should be kept at a theme level. Should I remove the PR about aligning figures from `MyST-NB` then?

najuzilu:
Also, is there a `:target:` equivalent in MyST-NB? Just wandering because I'm compiling a list of all parameters supported for figures.

choldgraf:
Hmmm, I am not sure what `target` is :-)  so...maybe?

najuzilu:
Actually the target is set by default. I didn't realize that figures were "clickable".

PR #61, opened 2020-04-18
adding support for opengraph tags

choldgraf:
It's a little bit hacky because the `html_baseurl` *must* be specified for this to work, but gets the job done. Unfortunately there's no way to dynamically update the meta tags so that they'd work with relative URLs

closes https://github.com/ExecutableBookProject/cli/issues/110


choldgraf:
I *think* this works after a card debugger, but let's see if it works on the main site

PR #77, opened 2020-04-27
full screen button

choldgraf:


choldgraf:
This adds a "full screen" button per @AakashGfude's suggestion, see it here:

https://233-244496643-gh.circle-artifacts.com/0/html/index.html

One downside to this is that the full-screen won't persist *across* pages, which could be annoying. We could fix this by using TurboLinks though this might create unpredictable behavior with the javascript libraries

AakashGfude:
Wow. This looks good. Thanks, @choldgraf. 
Agreed, we will need to load all javascript libraries beforehand if making it a single-page application.

PR #78, opened 2020-04-27
removing button backgrounds and text

choldgraf:
This prototypes removing button backgrounds and text (text now shows on hover). What do people think about it? cc @AakashGfude @jstac @najuzilu @chrisjsewell

### DIFF
--- a/sphinx_book_theme/static/sphinx-book-theme.css
+++ b/sphinx_book_theme/static/sphinx-book-theme.css
@@ -189,7 +189,7 @@ dl.footnote dd p {
   .topbar > div {
     height: 2.5em;
     top: 0px; }
-  .topbar .topbar-main button.navbar-toggler, .topbar .topbar-main .dropdown-buttons-trigger {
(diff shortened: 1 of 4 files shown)


choldgraf:
I think that it looks cleaner - if folks don't think that removing the text is a problem, then I'd be +1 on merging this and iterating from there...

AakashGfude:
+1 on this. It looks neat. I also think that those tooltip texts `download this page` and `launch interactive content` might be a bit obtrusive for the dropdown and we can do without it?

jstac:
Yes, it looks much cleaner.  Nice one.  Without text there's less distraction from the heading.  

@chrisjsewell also suggested right aligned and that sounds good too --- but either way I agree on making this change and then iterating.

choldgraf:
let's merge this one and then we can iterate on right-aligned in another issue. I want to leave the tooltips for now, partially because they're a bit more accessibility-friendly. Thanks for the feedback!

PR #79, opened 2020-04-28
Buttons

choldgraf:
This addresses many of @chrisjsewell's points in https://github.com/ExecutableBookProject/cli/issues/129

It:

* Standardizes button size
* Uses bootstrap for the tooltips and chooses positions that don't overlap with other buttons
* Moves the buttons to the right of the page
* for "focus" mode, it's a bit tricky to get just the content to expand, but as a hack in this PR we toggle the sidebar when that button is clicked as well, so that the only thing people see is the topbar + the toc

Here's a screenshot:

image

and here's hover behavior:

what do folks think?


choldgraf:
I'm gonna go ahead and merge this because I think folks are +1 on all the changes here, we can iterate on these in subsequent issues/PRs

PR #90, opened 2020-05-15
fix wrong kernelspec

joergbrech:
Hey, @choldgraf,  the fix in this PR resolved https://github.com/executablebooks/jupyter-book/issues/584 for my octave jupyter book.


choldgraf:
@joergbrech amazing! thanks for the fix :-)  do you think you could add a test for this? I think if we added a notebook w/ a kernel that wasn't python3 here: https://github.com/executablebooks/sphinx-book-theme/tree/master/tests/sites/base then we could test to make sure that the proper button link was created in `test_build`. WDYT?

joergbrech:
Sure! I will check it out next week.

joergbrech:
Ok, I found some free time today :smile:

choldgraf:
amazing! Thanks very much for the fix :-)

PR #91, opened 2020-05-15
Added ul and li style for toc directive

AakashGfude:
@choldgraf  styling for the list headers of {tableofcontents} directive. What do you think?


choldgraf:
Hmmm - this is going to annoy you, and I apologize :-)  but, if we're going to put styles in this theme for the TOC directive, why don't we move the directive (and tests) here instead of in jupyter-book? Otherwise there's no way for us to test out the styling here without also trying it out in jupyter book, which feels a bit clunky. What do you think?

AakashGfude:
Hmm, it makes sense in a way. But, I am not sure if this functionality should be inside a theme, as it does not add structure or design to the page. Will it be obvious to people using ebp to find this directive here? no need to apologize though :)

choldgraf:
fair enough - though I just realized that the styles in this PR won't actually be applied because I don't believe that the table of contents directive uses those classes, no? Will that need to be a follow-up PR?

I just opened up https://github.com/executablebooks/jupyter-book/pull/597 so we can demo the use of the tableofcontents directive and see how it looks

AakashGfude:
@choldgraf  Great. Thanks for the PR. 
tableofcontents uses those classes. I add the classes to the node during building the node tree.

choldgraf:
Hmmm - the tableofcontents isn't showing up in the PR preview, any idea why that would be? https://deploy-preview-597--jupyter-book.netlify.app/content-types/index.html

AakashGfude:
@choldgraf apparently, the `self.env.titles` variable is not fully constructed at this point. It has some files but not all. the value of the variable was something like :- 
image



So, the line of code here https://github.com/choldgraf/jupyter-book/blob/docsimprove/jupyter_book/directive/toc.py#L64 , just exits the loop as it does not find an entry for the files. 
Do you know any reliable way in sphinx to fetch the titles of the document from there path entry in _toc.yml.

When you insert a debugger, and then after that rebuild it without it, then it works . So probably did not see it because of  all the debuggers in place.  Weird

choldgraf:
> Do you know any reliable way in sphinx to fetch the titles of the document from there path entry in _toc.yml

Hmm good question - I guess you could load the doctree itself and grab the first `title` element, I think that should be the title in a reliable way. I'm not sure if there's a special Sphinx function for it...

AakashGfude:
Thanks. Actually, I got the idea of using this env variable from `TocTree` class only, like here https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/environment/adapters/toctree.py#L142 . But they have more complex machinery around it then me. I will go through it properly then, later in the day when I have time.

choldgraf:
One thought I just had: I wonder if we could reduce some complexity by actually using the toctree directive that is added in toc.py... Currently it adds toctree but with a hidden flag. What if, before doing that we just did a search for tableofcontents directives and it found, replaced it with the toctree, *without* the hidden tag

AakashGfude:
hmm. we can do that. But, when I was trying to use toctree directive earlier under the hood for this, I was getting circular dependencies in some cases and ultimately maximum recursion depth exceeded error. But in that case, I was using the toctree node programatically instead of wrting it in md.
I guess this error will not come in your case, or maybe it will come once we unhide it? not sure.
But, also in a way it feels like we will actually be increasing complexity if we use something complex like a toctree for this.
Anyways, will have to fiddle with it and see how it works out.

choldgraf:
Though in that case you were creating *another* toctree object, I believe, right?

I'm suggesting that here, where we insert the toctree to the end of each page: https://github.com/executablebooks/jupyter-book/blob/master/jupyter_book/toc.py#L141

we can either simply remove the `:hidden:` flag here: https://github.com/executablebooks/jupyter-book/blob/master/jupyter_book/toc.py#L71 (perhaps with a `show=true` argument in `_toc.yml`) or, if somebody wants to choose the *location* of the toctree, they can insert a `{tableofcontents}` directive in the page's content somewhere, and the toctree will be inserted there instead of at the end

AakashGfude:
right. I will have a go at this.

PR #101, opened 2020-05-23
Fixing small typo: missing format character

nathancarter:
Otherwise the error would contain the literal text {notebook_interface}, rather than its meaning.


choldgraf:
good catch, thanks 👍

PR #103, opened 2020-05-26
only hiding the right TOC if it overlaps with sidebar content

choldgraf:
This now calculates the height of the TOC, and hides it only if an element with a class that is associated with "right margin" space overlaps with the sidebar.


choldgraf:
@chrisjsewell this should satisfy the condition that people use *no margin content* and want their TOC to show up all the time 👍

PR #111, opened 2020-05-29
fixing suffixes

choldgraf:


choldgraf:
closing because this is actually a bug in JB

PR #123, opened 2020-06-19
Feature/add colab launch button

### LINKED ISSUE #107: Permit Colab links in addition to or instead of Binder links
It is nice and easy to turn any Jupyter notebook on GitHub into an executable one on Colab by simply visiting any URL of the form `https://colab.research.google.com/github/USER/REPO/blob/master/PATH/FILENAME.ipynb`.  This is often more desirable than Binder because of the much faster load time of the Colab environment than the Binder one, and of course the ability to save your own copies of notebooks.  Could this be an option in jupyter-book, at least for those books published to GitHub?

lewtun:
* Add launch button for Google Colab
Screen Shot 2020-06-19 at 10 52 41 am
* Add file regression test for launch buttons. This supersedes the unit test for launch button URLs and thus the assertions in `test_build_book` are removed.
* Remove JupyterHub launch button from docs configuration since it points to a Berkeley server that is not accessible by the public. Note that we keep the JupyterHub launch button in the test configuration since we wish to test the HTML itself is correct (not the accesssibility of a valid URL).

Closes #107


tests/test_build.py
```diff
@@ -59,6 +59,10 @@ def test_build_book(file_regression, sphinx_build):
             f"https://datahub.berkeley.edu/hub/user-redirect/git-pull?repo=https://github.com/executablebooks/sphinx-book-theme&urlpath=lab/tree/sphinx-book-theme/TESTPATH/{path}"  # noqa E501
             in ntbk_text
         )
+        assert (
```
choldgraf, line 62: I think that we can remove all of the `assert` statements now that you're using the regressions test

lewtun: done!

lewtun:
Thanks for the feedback @choldgraf! I've made the changes you suggested and updated the description of the PR accordingly. I'm happy to take care of the tweak to Jupyter Book once the next release of SBT is out.

choldgraf:
nice! this is looking good to me! let's give it a merge and then get it into Jupyter Book! I'll cut a SBT release now.

I just double-checked and in Jupyter Book, the config should automatically work (https://github.com/executablebooks/jupyter-book/blob/master/jupyter_book/yaml.py#L30). So, the only PR that's needed there is to document this feature! Wanna do that?

Thanks for the improvement! 🎉

lewtun:
sure, happy to take care of the docs in Jupyter Book! thanks for helping out with the review 😃

PR #125, opened 2020-06-24
using sphinx-thebelab

choldgraf:
This removes thebelab-specific JS from this repository and uses `sphinx-thebelab` instead.

cc @akhmerov - here's an example of how we can use `sphinx-thebelab` to build on top of it, WDYT


akhmerov:
Looks good!

Shouldn't configuring the extension this also go into theme itself? Every user of the book theme will have the same selectors.

choldgraf:
@akhmerov yeah I was thinking that too. We could just "update" the `sphinx-thebelab` configuration with whatever is in the sphinx-book-theme?

akhmerov:
Exactly, that's what I was thinking as well.

choldgraf:
@akhmerov just made some pushes to update the UI and configurability. What do you think?

(note taht I also realized that the binder environment doesn't actually work with the code that's here 😬 but that's fixed in this PR, just won't work until it's merged)

choldgraf:
gonna merge this so that we can try it out and see how it works

akhmerov:
Sorry, late to the party. Everything looks good at a glance.

The only things I noticed were:

- The branch could be potentially obtained from git, as opposed to being `master`
- `thebe_config.get("selector_input", "") + ",.cell_input div.highlight"` could seemingly produce an illegal selector if the first part is empty. Shouldn't this be `','.join(...)`?

choldgraf:
ah that's why I'm `lstrip`ing everything just after that :-)

re: the branch obtained from git...could you clarify a bit?

akhmerov:
e.g. using the output of `git rev-parse --abbrev-ref HEAD` instead of `master`. Not sure whether it's a reasonable default: it would slow down binder on PRs due to rebuilding the image, but would allow testing easier.

choldgraf:
maybe we should re-name "branch" to "ref"? Then it could point to commits etc

akhmerov:
I didn't mean the name of the variable as much as actually doing `subprocess.check_output` to see whether we are in a git repository and whether we can infer a good branch name. Also not hard-coding `master` is maybe a good idea with the discussions of alternative default branch names.

choldgraf:
ahhh I see what you mean - so you mean if no explicit branch is given, then try to infer the branch from the current repository, and if that doesn't work then defer to `master`?

akhmerov:
Yes, exactly!

choldgraf:
it's a cool idea - wanna open up an issue about it?

akhmerov:
Sounds good, I was just wondering: should it be here or in sphinx-thebelab?

choldgraf:
good point - probably sphinx-thebelab

PR #133, opened 2020-07-17
Fixed font scaling for caption spans

### LINKED ISSUE #131: Awkward handling of mathjax in figure captions
I'm seeing some awkward clipping behavior on mathjax embeds in Figure captions.  In Firefox, I see this:
image

which is generated by the following markdown:

````
```{glue:figure} shifting

*Left*: an impulse (top), a negative impulse (middle), and a 3-step delay (bottom).
*Middle*: the DFT spectrum (real and imaginary) components of each signal are distinct.
*Right*: all three signals have identical magnitude spectra $|X[m]| = 1$.
```
````
Note the truncated math span for `$|X[m]| = 1$`.

In Chromium, I see this:
image
which is at least not truncated, but the scaling seems incorrect to me.

(NB: I have some custom CSS to do justified text layout, but that's independent of this problem.)

The problem appears to come from a bad interaction between the book theme and mathjax due to setting `font-size: 0.9em` here https://github.com/executablebooks/sphinx-book-theme/blob/3d27feb9c375b2f63fa54e24709a0cecdbe40d58/sphinx_book_theme/static/sphinx-book-theme.scss#L121-L123

This appears to be fixed if I hack this in the console to scale up to `font-size: 0.9rem` (root scaling).  It then renders as (in Firefox):
image

and in Chromium:
image

In both cases, the surrounding typesetting seems .... broken (note the unpredictable placement of the following punctuation).  I'm not enough of a CSS wiz to sort that one out.  **EDIT**: I think this one might be down to mathjax.

---
This problem may pop up in other places, and it's probably worth doing a thorough audit to ensure that root scalings are used where appropriate.

bmcfee:
This PR fixes #131 by using root-scaling (rem) on caption spans.

It looks like the weird behavior I was seeing from mathjax was an artifact of hacking this behavior in the browser via the CSS editor after the math had already been rendered.  Having now fixed the scaling in the source css, mathjax behaves properly in both firefox and chromium for me.

### DIFF
--- a/sphinx_book_theme/static/sphinx-book-theme.scss
+++ b/sphinx_book_theme/static/sphinx-book-theme.scss
@@ -119,7 +119,7 @@ main.bd-content {
             }
 
             span {
-                font-size: .9em;
+                font-size: .9rem;
             }
         }
 


choldgraf:
thanks for the fix! I must admit I still am confused by the different between `em` and `rem` :-P

let's :shipit: !

bmcfee:
Thanks!  

BTW: i didn't thoroughly audit for other places that might be affected by this, but i'm not seeing it pop up in my book build after fixing.  I guess just keep an eye out for bad mathjax cropping / scaling in any block that implements a font scaling.

choldgraf:
yeah - I still have no idea how to "test" things like CSS bugs beyond the "are github issues opening about it" test :-P

PR #140, opened 2020-07-29
👌 IMPROVE: Citation handling with config

AakashGfude:
The default behaviour where `superscript_citations=False` :- 

image

When `superscript_citations = True` :- 

image

**Working on adding tests**

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -64,6 +64,7 @@ html_theme_options = {
     "use_issues_button": True,
     "use_repository_button": True,
     "expand_sections": ["reference/index"],
+    "superscript_citations": False
     # For testing
     # "home_page_in_toc": True,
     # "single_page": True
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,8 +63,7 @@ html_theme_options = {
     "use_edit_page_button": True,
     "use_issues_button": True,
     "use_repository_button": True,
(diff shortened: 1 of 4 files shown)


choldgraf:
@AakashGfude wanna switch this one up so that we:

* don't use config for this
* but make the default CSS "non-superscript"
* add some docs in an "advanced" or in the "html" section about how to make the refs superscript

?

AakashGfude:
> * add some docs in an "advanced" or in the "html" section about how to make the refs superscript

Just wondering how to achieve this point. A normal user will only install everything though pip right, and won't really access the code base of the installed packages. 
To me, it feels like the only place he will have access to the stylesheet will be in the `html` folder of the `_build` folder, which is temporary as he might have clean builds.

choldgraf:
I was imagining we'd refer to https://jupyterbook.org/advanced/sphinx.html#custom-css-or-javascript and tell people what CSS rules they'd need to add...

AakashGfude:
aah. Thanks @choldgraf , did not know about this feature. Awesome. Still catching up on everything after this long break :sweat_smile:

choldgraf:
@AakashGfude is this ready to go? I assume you wanna add the docs for this in jupyter-book

choldgraf:
gonna merge this in so we can cut a release but feel free to open a new PR to clean stuff up 👍

PR #154, opened 2020-08-06
Drop left sidebar opacity change-on-hover behavior

### LINKED ISSUE #106: Book menu incorrectly displayed in iPad
**Describe the bug**
I have noticed a bug when navigating the following url from an iPad: [https://jupyterbook.org/intro.html](https://jupyterbook.org/intro.html).

The menu is displayed like grayed/transparent on top of the text, and it doesn't look good at all.

**To Reproduce**
You may navigate to the url above from an iPad, or just use chrome developer tools and select iPad.

I have copied below a screenshot showcasing the issue and how to reproduce it.

Screenshot 2020-05-27 at 16 36 19

Hope it helps!

pradyunsg:
Closes #106

### DIFF
--- a/sphinx_book_theme/static/sphinx-book-theme.scss
+++ b/sphinx_book_theme/static/sphinx-book-theme.scss
@@ -601,10 +601,6 @@ button.topbarbtn img {
         background: #c1c1c1;
     }
 
-    @media (min-width: $breakpoint-md) {
-        opacity: 0.6;
-    }
-
     // Hide the scrollbar until hover on wide screens
     @media (min-width: $breakpoint-lg) {
         &:not(:hover) {
@@ -618,10 +614,6 @@ button.topbarbtn img {
               background: #FFFFFF;
             }
         }
-    }
-
-    &:hover {
-        opacity: 1;
     }
 
     h1.site-logo {


pradyunsg:
Thoughts on this?

choldgraf:
I'm a little bit hesitant to do this, as I (personally) like the left sidebar opacity stuff. It is kind-of related to #160 - the reason that the fonts were smaller and in some cases greyed out in the left / right sidebar was in order to really drive the reader's focus to the content on the page. I find that having all of the words be the same font size across those UI elements makes each of them less-distinguishable, and un-greying the left sidebar would be another step in that direction :-/

I think the greying out does have some sub-optimal effects too (most notable for accessibility, I think)...are there other ways that we can more naturally separate the left/right sidebars from the main content and avoid the page looking like a wall of text?

pradyunsg:
Would it be OK to grey out the text instead of using opacity here?

choldgraf:
sure, that works fine for me as well...opacity just seemed like the easiest way to apply this across the whole sidebar (including the logo, headers, etc)

pradyunsg:
OK to merge this, and I file a follow up, or... do you prefer the grey-out-text-unless-hover stuff to be added it this PR?

choldgraf:
Let's merge and iterate 😊

Issue #144, opened 2020-08-06
Design: Don't use different font size for sidebar content


pradyunsg:
**Is your feature request related to a problem? Please describe.**

Font sizing changes in the sidebars make it difficult to read/skim the sidebar's contents and well as make it look out-of-place.

**Describe the solution you'd like**

Use the same size as the content for *all* navigation items in both sidebars.

- The indentation is sufficient to make the "depth" clear.
- Whitespace + bold text can be used for captions and toctrees to separate them visually.

See [readme.com's design for inspiration/idea](https://gist.github.com/pradyunsg/eeb74ac5b580d502c3d0e248310f273f#readmecom).

**Describe alternatives you've considered**

I can't think of any good ones. :)

**Additional context**

I'm happy to try implementing this myself, if there's interest in this idea. :)

pradyunsg:
FWIW, I did manage to draft a few changes toward this. Here are two screenshots (side-by-side) of the before and after.

<img width="226" alt="Screenshot 2020-08-06 at 6 25 02 PM" src="https://user-images.githubusercontent.com/3275593/89534178-2af60a80-d812-11ea-9ff8-3616ae512495.png"><img width="226" alt="Screenshot 2020-08-06 at 6 34 12 PM" src="https://user-images.githubusercontent.com/3275593/89535032-76f57f00-d813-11ea-9a2a-773613196a02.png">

And here's the two full-page demos:

(after)

<img width="1091" alt="Screenshot 2020-08-06 at 6 40 34 PM" src="https://user-images.githubusercontent.com/3275593/89535689-62fe4d00-d814-11ea-97b7-9567ad6afa34.png">

(before)

<img width="1091" alt="Screenshot 2020-08-06 at 6 40 52 PM" src="https://user-images.githubusercontent.com/3275593/89535684-609bf300-d814-11ea-8717-0a53592ca784.png">

chrisjsewell:
Well you have been busy @pradyunsg 😆 
Looks great to me!

choldgraf:
+1 from me, I think that's reasonable and will help with accessibility efforts

choldgraf:
oh wait, I see you've already opened up a PR for this which I missed because **I have like 50 github notifications from you** ;-)

chrisjsewell:
I think this is now closed in #156, but feel free to re-open if not

PR #156, opened 2020-08-06
👌 IMPROVE: Consistent font size in navigation

pradyunsg:
Toward #144

### DIFF
--- a/sphinx_book_theme/static/sphinx-book-theme.scss
+++ b/sphinx_book_theme/static/sphinx-book-theme.scss
@@ -741,12 +741,11 @@ nav.bd-links {
         width: 100%;
     }
 
-    li.toctree-l1 {
-        font-size: 1.2em
-    }
-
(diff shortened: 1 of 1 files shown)


choldgraf:
I think this generally looks good. Though the sidebar feels a little bit cramped now that they're all the same size. I wonder what you think about slightly growing the vertical space between top-level left sidebar sections. So something like:

image

(which I added with this rule)

```css
ul.sidenav_l1 > li {
    margin-bottom: 0.2em;
}
```

pradyunsg:
Definitely! Lemme wrap up dinner and I'll add the .2em to the elements. I *think* we already set margin/padding on li elements, so this should fit right in.

pradyunsg:
Done!

choldgraf:
nice - OK I'm gonna merge this, though perhaps these changes will also relate to the conversation in #160 and we can always iterate further

choldgraf:
thanks @pradyunsg !! 🎉

PR #157, opened 2020-08-06
👌 IMPROVE: Use consistent font size in toc

pradyunsg:
Toward #144

### DIFF
--- a/sphinx_book_theme/static/sphinx-book-theme.scss
+++ b/sphinx_book_theme/static/sphinx-book-theme.scss
@@ -659,6 +659,10 @@ button.topbarbtn img {
     }
 }
 
+.toc-h2, .toc-h3, .toc-h4 {
+    font-size: inherit !important;
+}
+
 .site-navigation, .site-navigation.collapsing {
     transition: flex .2s ease 0s, height .35s ease, opacity 0.2s ease;
 }


choldgraf:
The main reason that this was a bit smaller was:

1. To make it more distinguishable from the main content
2. To avoid it getting cramped in the (relatively small) amount of space available to it

I'm open to bumping up the font size but curious what you think about those two concerns

pradyunsg:
Re 1: The positioning (indentation relative to content around + place on page) do a good enough job of distinguishing it IMO.

Re 2: This is *definitely* a valid concern.

There's a bunch of indirect-effects to balance (nudges for title length / less deeper headings, legibility of content etc) but I think it comes down to a choice of tradeoffs. The question I'd asked myself was: which is better, having h4 be more difficult to read (because of small font size) or need wrapping when it's moderately long? My answer was the latter.

TBH, what motivated this entire exercise was that I couldn't read the h4 easily without moving closer to the screen and wanted to try fixing it because it seemed easy enough. :)

choldgraf:
> what motivated this entire exercise was that I couldn't read the h4 easily without moving closer to the screen

I think if you are having this experience then others are as well, so we should merge this

choldgraf:
thanks for the improvement! 🎉

PR #168, opened 2020-08-09
🔧 MAINT: right tocbar font size

choldgraf:
sets the right contents bar to have the same font size as the left navigation bar. @pradyunsg what do you think?


choldgraf:
yeah good point for "just in case" - right now it won't show up if there's only one title on the page

PR #171, opened 2020-08-10
🐛 BUG: backref superscript

AakashGfude:
The following PR makes the backref inline instead of superscript.

cc @choldgraf 

fixes https://github.com/executablebooks/MyST-Parser/issues/196


choldgraf:
Definitely an improvement - it's a little lower than the text to the left, maybe we can just remove the "vertical-alignment" rule alltogether and it'll line up?

AakashGfude:
good eye @choldgraf . have removed it.

choldgraf:
woo, looks great!

Issue #153, opened 2020-08-06
Add a shadow to the top bar on scroll


pradyunsg:
**Is your feature request related to a problem? Please describe.**

Content at the top of the page gets cropped, seemingly for no reason.

**Describe the solution you'd like**

A shadow underneath the top bar would make it easier to perceive that there is an element that's supposed to be on top, making it clearer why content is getting cropped (there's something above it!).

**Describe alternatives you've considered**

None.

**Additional context**

See https://squidfunk.github.io/mkdocs-material/getting-started/ for a decent+subtle implementation of this.

choldgraf:
very cool - I like that idea 👍

pradyunsg:
The top bar is... interestingly styled. Would anyone mind if I rework that element's HTML?

choldgraf:
Ummm, depends on what your plastic surgery looks like when it's done :-)

choldgraf:
If the end result behaves the same then in general I'm +1 though

pradyunsg:
How do folks feel about me taking this and #148 down in one big PR?

I'm gonna need to make changes to the sidebar for this, and vice versa, so I'm thinking to kill 2 issues with one PR.

chrisjsewell:
yep if its easier. we can always split it later if its looking too unruly

choldgraf:
go for it (with the caveat that the more time goes on, the greater chance I will drop off the face of the earth for a bit)

Issue #148, opened 2020-08-06
Make left sidebar into an off-canvas drawer on mobile/tablet


pradyunsg:
**Is your feature request related to a problem? Please describe.**

- The (left) sidebar "comes in from the top" of the page, through resizing of the corresponding element. This implementation requires JS to work [^1] and shows up in a different place than the desktop sidebar.
- The (left) sidebar does not degrade gracefully when JS is disabled. A user who has JS disabled for some reason (*ahem* like me) cannot view documentation built with this theme on mobile, since the sidebar cannot be collapsed and covers regular content. And they can't access the navigation either.

**A link to the documentation page where you see an issue.**

Any page works TBH. Here's the page I have open right now: https://sphinx-book-theme.readthedocs.io/en/latest/index.html

**Describe the solution you'd like**

I suggest we switch to the same mechanism+design for providing the sidebar as the mkdocs-material theme: https://squidfunk.github.io/mkdocs-material/getting-started/. Try it on ~1024px wide windows, which shows how it works best IMO. That mechanism works without needing JS (it's pure-CSS) and, thus, degrades gracefully.

Note that I'm not suggesting we change the contents/design of what's in the sidebar in this issue, but only how it is handled in terms of being visible/not visible.

FWIW, I have experience with implementing this style of a sidebar, from when I was working on re-implementing mkdocs-material for Sphinx [here](https://github.com/pradyunsg/materialism). I'm happy to try re-implementing it in the context of this theme, if there's interest in making this switch.

pradyunsg:
Oops, I used the wrong template.

choldgraf:
Totally I agree, I think this is a more common and probably better pattern here. To be brutally honest the only reason that the sidebar comes down from the top is because I spent two days trying to figure out how to do it from the left and couldn't find the right CSS / JS incantation to do it in bootstrap 🤦‍♂️

chrisjsewell:
Yep definitely. You're beating me to features I wanted to look into @pradyunsg, which I absolutely have no problem with 😄

mmcky:
@DrDrij would you be able o take a look at this `feature` request?

choldgraf:
I believe that in the mock-up that @DrDrij showed a few months back he had some nice modifications made to the left sidebar, would love to see some of those in code form as well if you’re thinking of tackling this one too!

DrDrij:
[...]

PR #172, opened 2020-08-10
👌 IMPROVE: Nicer Development Experience

pradyunsg:
Closes #151 in my opinion. :)

Following @choldgraf's [nudge](https://github.com/executablebooks/sphinx-book-theme/issues/151#issuecomment-670637019), I ended up writing the entire noxfile for this project *and* rewrote a part of the contributing guide to reflect what the various commands needed during development look like with nox.

Oh, and I added the `nox -s docs-live` thing, with live-reload through sphinx-autobuild as well. :)

PR #177, opened 2020-08-12
Nicer Development Experience, take 2

pradyunsg:
A couple of smaller changes, based off experience of working on #148 and #153.


chrisjsewell:
What were the issues you come across with the current setup?
I guess the only issue with rebuilding the whole documentation every time (I think thats what `-a` does?) is that its a pain when you're just playing around with SCSS.
Not so much for the current documentation here (because it's only a few pages), but thinking towards using this approach for messing around with larger sets of documentation.

pradyunsg:
- coveralls broke
- changes to layout.html only get reflected on index (because that's the only page sphinx detected as outdated as per our extension
- including the generated css in the repository makes it significantly harder to cherry-pick specific changes (unless we wanna un-compress the generated css, which I don't)
- sdist -> wheel doesn't work right now, IIUC.

pradyunsg:
To be clear, I'm pretty sure I broke all these things in #172. 😅

chrisjsewell:
😂 you gotta break some eggs to make an omelette

pradyunsg:
Everything except coverall should be working. Let's do coveralls in a follow up PR. :)

sphinx_book_theme/__init__.py
```diff
@@ -22,6 +22,13 @@ def add_static_path(app):
     static_path = Path(__file__).parent.joinpath("static").absolute()
     app.config.html_static_path.append(str(static_path))
 
+    # Compile the css file if it's not been compiled already
+    compiled_css_file = static_path / "sphinx-book-theme.css"
+    if not compiled_css_file.exists():
```
chrisjsewell, line 27:
```suggestion
    compiled_css_file = static_path / "sphinx-book-theme.css"
    compiled_css_file2 = static_path / "_print.css"
    if not (compiled_css_file.exists() and compiled_css_file2.exists()):
```

just to be safe?

pradyunsg: That file won't get compiled into a css. That's what the `_` does. FWIW, I am borderline wondering if I should break out the main scss file into small `_sidebar` and friends. :)

pradyunsg: Relevant Stack-Overflow question to save you from a round of web searches: https://stackoverflow.com/questions/34889962/why-put-in-front-of-the-file-name-or-in-scss-css

chrisjsewell: Ah cheers; just showing by "newbness" on SCSS

chrisjsewell:
Yep I think if you push another commit codecov should be working again now

pradyunsg:
You underestimate how lazy I am. 🙃

chrisjsewell:
Ah turns out the codecov app wasn't enabled yet for sphinx-book-theme (this is done in the organisation setings)

chrisjsewell:
@pradyunsg why did you break nox lol:

```
(base) markdown-it-py/$ nox                     
nox > Running session tests
nox > Creating virtual environment (virtualenv) using python3.7 in .nox/tests
nox > Command /Users/chrisjsewell/opt/miniconda3/bin/python3.7 -m virtualenv /Users/chrisjsewell/Documents/GitHub/markdown-it-py/.nox/tests failed with exit code 1:
ImportError: cannot import name 'DETACHED_PROCESS' from 'virtualenv.util.subprocess' (/Users/chrisjsewell/opt/miniconda3/lib/python3.7/site-packages/virtualenv/util/subprocess/__init__.py)
```

https://github.com/pypa/virtualenv/commit/0cd009b5a1338f66397f71c85a75f576a2f3eabf

pradyunsg:
Any blockers to merging this? :)

sphinx_book_theme/__init__.py
choldgraf: So does this mean that if we change the scss file we should delete the css file so it is recompiled?

pradyunsg: You should if you're using `nox -s docs-live` or `nox -s docs`, which do recompile the SCSS.

The only thing this exists for, is to be able to install+use this package off VCS. :)

choldgraf:
I just tried the nox workflow and it seems to be taking forever to create the virtualenv. It's been sitting on `pip install .[sphinx]` for about 5 minutes now. Is this expected? Is there any gotcha w/ nox that I'm missing? (I am on WSL if that matters)

chrisjsewell:
FYI, I haven't had any issues with this; you've gotta expect on Windows, that everything's gonna be rubbush 😆 
but yeh we should probably put the makefile back as an alternative

pradyunsg:
> for about 5 minutes now. Is this expected?

Do you have a virtualenv in the same folder? If so, you're probably hitting https://github.com/pypa/pip/issues/2195, which... well, needs fixing. :)

>  but yeh we should probably put the makefile back as an alternative

Separate PR incoming.

chrisjsewell:
@pradyunsg FYI @choldgraf is away for a bit now. I'm meaning to get round to properly looking at this soon (and sphinx-book-theme in general), but let me know if not merging is holding you up in any way on any of the other improvements

PR #179, opened 2020-08-13
Let's fix codecov

pradyunsg:
Broke in #172. Built on top of #177.

Let's see if I can figure out *why*, because `pytest` does generate the file codecov wants to use, when I run pytest locally.


chrisjsewell:
When you say coveralls, I guess you mean codecov?

pradyunsg:
lol

pradyunsg:
The file exists. Why is codecov not happy with it? >.<

https://github.com/executablebooks/sphinx-book-theme/pull/179/checks?check_run_id=979013832#step:6:11

chrisjsewell:
note codecov seems to be not working in all repositories at the moment

choldgraf:
Maybe this is a setting in codecov itself we need to change? Could it have been broken back when we renamed the org?

chrisjsewell:
As I m motioned somewhere else, I think codecov is playing up in general in the last ~24 hours

chrisjsewell:
The checks aren't showing up in any PRs

pradyunsg:
So... you're saying... that I didn't break it?

\o/

chrisjsewell:
> So... you're saying... that I didn't break it?

Oh no, I still blame you for breaking all of CodeCov 😆

chrisjsewell:
yep codecov is reporting again on workflow re-run, you might have to push a commit to get it to show up in the checks?

pradyunsg:
Nothing to do here, it seems. :)

PR #190, opened 2020-08-23
🐛 FIX: checking for title in first page

choldgraf:
closes https://github.com/executablebooks/sphinx-book-theme/pull/190

### DIFF
--- a/sphinx_book_theme/__init__.py
+++ b/sphinx_book_theme/__init__.py
@@ -83,7 +83,10 @@ def add_to_context(app, pagename, templatename, context, doctree):
             master_doc = config["master_doc"]
             master_doctree = app.env.get_doctree(master_doc)
             master_url = context["pathto"](master_doc)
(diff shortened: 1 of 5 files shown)


choldgraf:
FYI @chrisjsewell I haven't merged this because I can't figure out how to write the proper test for it. I may not have much time over the next few weeks to get to it. So I guess we could merge this in w/o a test but I'm hesitant to do so

chrisjsewell:
Yeh no worries, I'll try to have a look at some point

choldgraf:
Because another issue got opened that relates to this (https://github.com/executablebooks/jupyter-book/issues/966) I tried to another stab at it.

I've added a test that passes locally, but I can't figure out why it's not passing on here. (I'm not able to use nox yet because it's still freezing on me when I try to install it)

@chrisjsewell do you have any ideas?

choldgraf:
ok got this working after changing our pytest fixture scope

PR #191, opened 2020-08-24
🐛 FIX: Add formatting for definition/field/apidoc lists

### LINKED ISSUE #158: No CSS for general definition (dl,dt,dd) tags
See https://www.w3schools.com/tags/tag_dl.asp

This means, for example, sphinx API documentation looks rubbish, because there is no delineation between components, e.g.

- RTD theme: https://ipypublish.readthedocs.io/en/latest/api/ipypublish.convert.main.html
- book theme: https://markdown-it-py.readthedocs.io/en/latest/api/markdown_it.main.html

chrisjsewell:
closes #158 

The CSS for the API doc components is taken from the RTD theme CSS, which I converted to SCSS here: https://gist.github.com/chrisjsewell/0f414e6c13e714f9cf70c36f20efaafa
There can possibly be some "pruning" of unneeded elements?

As noted in the scss, field list should be a different format, but that format is a little tricky (see https://github.com/pandas-dev/pydata-sphinx-theme/issues/193) and not widely used, so I think tackle that in a separate issue/PR

@pradyunsg I also added nitpick mode (to check for missing refs) to the nox sphinx-build commands


chrisjsewell:
Ok merge this and make a release, because I need it for myst-parser

PR #200, opened 2020-08-31
⬆️ UPGRADE: Use pyScss instead of libsass

### LINKED ISSUE #195: Replace libsass with pure python implementation
I want to create a docker image based on alpine linux for sphinx and this awesome theme. Libsass package is too complicated, it must be compiled and it is too heavy (~60MB) also. Would it be possible to replace it with another written in pure python? For example https://github.com/Kronuz/pyScss?

hason:
Fix #195

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
         "sphinx",
         "click",
         "setuptools",
-        "libsass",
+        "pyScss",
         "pydata-sphinx-theme~=0.3.0",
         "beautifulsoup4",
     ],
--- a/sphinx_book_theme/__init__.py
+++ b/sphinx_book_theme/__init__.py
@@ -4,7 +4,7 @@ from docutils.parsers.rst import directives
 from docutils import nodes
 from sphinx.util import logging
 from bs4 import BeautifulSoup as bs
-from sass import compile as sass_compile
+from scss.compiler import compile_file
 
 from .launch import add_hub_urls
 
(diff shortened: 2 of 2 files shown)


chrisjsewell (approved):

chrisjsewell:
Well I'm happy to merge this. @pradyunsg any objections?

pradyunsg:
As long as pyscss is a drop-in-ish replacement to libsass in terms of syntax etc, no objections. :)

pradyunsg:
And, thanks for checking! ^>^

chrisjsewell:
> And, thanks for checking! ^>^

Well its always nice to be kept in the loop. Which is apparently not part of the conda-forge policy 👀 https://github.com/conda-forge/staged-recipes/pull/12511

choldgraf:
+1 from me if it's more lightweight and the outputs are the same (which they seem to be)

chrisjsewell:
goddamit, straight after I merge, I find out that pyScss doesn't seem to support `@use` and only the defunct `@import` https://sass-lang.com/documentation/at-rules/import

chrisjsewell:
linking https://github.com/Kronuz/pyScss/issues/409

chrisjsewell:
Err, I'm not sure about this compiler, now I start to see unadressed issues like this https://github.com/Kronuz/pyScss/issues/370

chrisjsewell:
hmm, libsass doesn't see to support it either, or at least I can't get it working?
Seems strange that there's no support, when there is such a big warning sign about `@import`

chrisjsewell:
Hmm, libsass doesn't seem to support it either? Or maybe its just me being a newb

chrisjsewell:
it also doesn't support sourcemap generation https://github.com/Kronuz/pyScss/issues/224

pradyunsg:
:(

Are we OK with the GitHub repository not being directly-installable?

chrisjsewell:
This is kind of funny; I created an issue about the maintenance status https://github.com/Kronuz/pyScss/issues/410, mentioning a few PRs that had been open for months, and within a few hours 4 PRs had been merged of which I was thanked for one.

Its still an annoyance that source maps aren't available

> Are we OK with the GitHub repository not being directly-installable?

I have no strong opinions on this, I rarely install directly from a repo.
You probably know more than most if this is an issue?

PR #201, opened 2020-08-31
Add sphinx extensions to docs/requirements.txt

najuzilu:
This PR adds the following Sphinx extensions under `docs/requirements.txt`:
* myst-nb
* sphinx-copybutton
* sphinxcontrib-bibtex
* sphinx-thebe

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,7 @@
 matplotlib
 numpy
 folium
+myst-nb
+sphinx-copybutton
+sphinxcontrib-bibtex
+sphinx-thebe


chrisjsewell:
I'm not sure if this is a good idea, at least for myst-nb, because now we are showing CSS that may have come from these extensions, rather than purely the CSS of the theme

chrisjsewell:
Actually all of these are in the `sphinx` extra anyway. Why do we need `doc/requirements.txt`, just for binder? Either we should make binder work with extras (the better option) or use only the file and remove the extra, since now they are duplicating each other

najuzilu:
Ah, I missed the `extras_require`. I assumed all the extras were in `requirements.txt` and  I kept running into `no module named ...` which is why I included them.

I'm happy with either as long as I don't have to manually install them haha.

pradyunsg:
Yea, you can run `pip install .[sphinx]` to get them. I think I'll go ahead and close this now, but do holler if this doesn't do the right thing for ya. :)

PR #211, opened 2020-09-22
♻️ REFACTOR: Asset compilation and dev workflow

chrisjsewell:
This refactor follows development of the workflow in https://github.com/executablebooks/sphinx-panels and https://github.com/executablebooks/web-compile.
It fixes a number of pressing issues (particularly see https://github.com/executablebooks/jupyter-book/issues/921):

- The package is using a SCSS compiler (pyScss) which is not well maintained and has some important un-adressed issues (e.g. it can't handle compiling CSS custom variables)
- Currently we are writing CSS files into the installed folder (not good at all!)
- The CSS/JS is not hashed, and so when people update their sphinx-book-theme, they may not always get the up-to-date files (due to caching)

The approach here is essentially very similar to pydata-sphinx-theme. But instead of using webpack to compile the SCSS/JS we are using the web-compile pre-commit hook. This limits the development complexity of having to work with an additional nodejs environment, and allows everything to run *via* tox.

As previously discussed, both of these approaches (including the compiled/hashed files in the repo) do suffer one drawback: that merge conflicts can arise if not making PRs to the latest commit on the `master` branch.
However, this use-case has literally never come up yet, so I'd rather accept this con for now, in return for the more important fixes/pros.
Note, there is talk of eventually doing this compilation at build time (see https://github.com/executablebooks/meta/issues/137), but I'm not satisfied yet that there is an ideal solution (and in any case this is a step in the direction).

As discussed previously I moved from nox to tox, since tox is (a) the more mature/used package and (b) used across other EBP packages.

The `tox -e docs-live`, in my testing, look to be working well, and now correctly updates as you change the source Python/SCSS/JS.

The contributing guide has also been updated to reflect these changes: https://sphinx-book-theme--211.org.readthedocs.build/en/211/contributing.html


chrisjsewell:
The contributing guide has also been updated to reflect these changes: https://sphinx-book-theme--211.org.readthedocs.build/en/211/contributing.html

docs/conf.py
```diff
@@ -41,6 +41,11 @@
 
 numfig = True
 
+myst_admonition_enable = True
+myst_deflist_enable = True
+myst_url_schemes = ("http", "https", "mailto")
+panels_add_boostrap_css = False
```
choldgraf, line 47:
```suggestion
panels_add_bootstrap_css = False
```

setup.py
```diff
@@ -55,25 +54,12 @@
...
+            "pytest~=6.0.1",
             "pytest-cov",
-            "beautifulsoup4",
-            "pytest-regressions",
+            "pytest-regressions~=2.0.1",
         ],
+        "live-dev": ["sphinx-autobuild", "web-compile~=0.2.1"],
     },
     entry_points={"sphinx.html_themes": ["sphinx_book_theme = sphinx_book_theme"]},
-    package_data={
```
choldgraf, line 65: this is unnecessary because we have the MANIFEST file right?

chrisjsewell: yep 👍

docs/contributing.md
```diff
@@ -20,96 +20,117 @@ git clone https://github.com/executablebooks/sphinx-book-theme
...
-If you're making changes to your local copy of this theme, there is a helper
-command :
+```console
+$ pre-commit run --all
+$ pre-commit install
+```
 
-```bash
-nox -s docs-live
+### Working on the theme
```
choldgraf, line 81: can we document `theme_dev_mode`

web-compile-config.yml
```diff
@@ -0,0 +1,13 @@
...
+      src/scss/sphinx-book-theme.scss: sphinx_book_theme/static/sphinx-book-theme.[hash].css
+    precision: 5
+    sourcemap: false
+    format: compressed
+  js:
+    files:
+      src/js/sphinx-book-theme.js: sphinx_book_theme/static/sphinx-book-theme.[hash].js
+  jinja:
+    files:
+      src/jinja/theme.conf.j2: sphinx_book_theme/theme.conf
```
choldgraf, line 13: so this pre-commit hook will generate the new CSS and JS files, + a new `theme.conf` file if the previous files have changed at all?

chrisjsewell: yep, give it a go! thats the best way to see how (if 🤞)  it works

chrisjsewell: and obviously you can have a look at https://github.com/executablebooks/web-compile if you want to understand further

docs/contributing.md
chrisjsewell: yeh, as mentioned in the code comments, this option should literally only be used in this repos conf.py; to allow for the CSS/JS to be dynamically updated

choldgraf:
hmmm, when I ran `tox -e py38-sphinx3`, it didn't seem to install most of the dependencies and then failed on `myst_parser is not installed`. Is there an extra command we're missing in the dev docs for installing the environment w/ tox?

chrisjsewell:
have you previously used tox in this folder, i.e. did a `.tox` folder already exist?
If so, have you tried either running: `tox -r -e py38-sphinx3`, or just deleting the folder.
If not, what is the console output?

choldgraf:
first time using tox, here's the output of the command you suggested:

```
py38-sphinx3 create: /home/choldgraf/github/forks/python/ebp/sphinx-book-theme/.tox/py38-sphinx3
py38-sphinx3 installdeps: sphinx>=3,<4
py38-sphinx3 develop-inst: /home/choldgraf/github/forks/python/ebp/sphinx-book-theme
py38-sphinx3 installed: alabaster==0.7.12,attrs==20.2.0,Babel==2.8.0,beautifulsoup4==4.9.1,certifi==2020.6.20,chardet==3.0.4,click==7.1.2,coverage==5.3,docutils==0.16,idna==2.10,imagesize==1.2.0,iniconfig==1.0.1,Jinja2==2.11.2,MarkupSafe==1.1.1,more-itertools==8.5.0,packaging==20.4,pluggy==0.13.1,py==1.9.0,pydata-sphinx-theme==0.4.0,Pygments==2.7.1,pyparsing==2.4.7,pytest==6.0.2,pytest-cov==2.10.1,pytest-datadir==1.3.1,pytest-regressions==2.0.1,pytz==2020.1,PyYAML==5.3.1,requests==2.24.0,six==1.15.0,snowballstemmer==2.0.0,soupsieve==2.0.1,Sphinx==3.2.1,-e git+https://github.com/choldgraf/sphinx-book-theme@e28bb5ce3d18dc8c3d00011ac224fd3894524301#egg=sphinx_book_theme,sphinxcontrib-applehelp==1.0.2,sphinxcontrib-devhelp==1.0.2,sphinxcontrib-htmlhelp==1.0.3,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.3,sphinxcontrib-serializinghtml==1.1.4,toml==0.10.1,urllib3==1.25.10
py38-sphinx3 run-test-pre: PYTHONHASHSEED='4177624336'
```

note that I don't see myst_parser etc being installed

chrisjsewell:
> Also, because I don't have Python 3.7 installed just running tox yields this error:

If you are using conda (?) you can just create an environment like `conda create -n python=3.7`, then`tox-conda` will be able to find it.

If not yeh you just have to run with the specified environment. You can set for `tox` to run all available python versions. But I've personally found that more hassle than its worth.

choldgraf:
hmmm, I installed `toc-conda` and while `tox` now does run, it has the same error as the manual tox command I showed above.

chrisjsewell:
> failed on myst_parser is not installed

Ah well thats because its only set to install the `testing` extras reuirements, and `myst-parer` is not part is not in that. Do we really need to install the whole `sphinx` extras to run the tests?

chrisjsewell:
but basically just replace `extras = testing` with

```
extras =
  testing
  sphinx
```

choldgraf:
well we do have a notebook and myst markdown files in the tests, so we'd at least need those two unless we want to re-write the tests to only use rST files

chrisjsewell:
Yeh but you don't really need to waste time installing numpy, pandas, matplotlib, etc.
But anyway just adding the the sphinx extra should fix it?

choldgraf:
yep adding `sphinx` got it working. I also tried adding `myst_nb` and `sphinx_thebe` to the tests requirements, and that got it working too...maybe that is the easiest path forward?

chrisjsewell:
yep thats what I was just doing 👍 Obviously the idea with these dev environments is that they should only contain what is actually necessary to perform the command

choldgraf:
cool, makes sense to me too, probably just cruft from the old days.

The pre-commit workflow worked for me 👍 once you get this merged I can give it a shot with a fix for https://github.com/executablebooks/jupyter-book/issues/986#issuecomment-696967157

setup.py
```diff
@@ -53,6 +53,10 @@
             "sphinx-thebe",
         ],
         "testing": [
+            "myst_nb~=0.10.1",
+            "sphinx_copybutton",
```
choldgraf, line 57: I _think_ we can remove both the copybutton/togglebutton from the tests (and the `conf.py` in the site that the tests build)

PR #214, opened 2020-09-23
✨ NEW: Add translations for all buttons and tooltips

### LINKED ISSUE #197: Support Sphinx language for some menu options
In Sphinx, we are able to detect the language being used at build time:

https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language

We could use this along with a dictionary that chooses the right words/phrases for certain UI elements in the documentation (e.g. "search" and "contents"). Even if these are customizable it would be better to use the same internationalization mechanisms that Sphinx has to use reasonable defaults.

chrisjsewell:
closes #197


chrisjsewell:
image

choldgraf:
nice - how did you generate these translations? Could we document either in the contributors docs how this is structured, or the user docs how they can leverage this functionality? I suspect many are not familiar with Sphinx translation infra

chrisjsewell:
ဟုတ်ပါတယ်၊ ငါဒီဘာသာစကားအားလုံးကိုပြောတယ်။

chrisjsewell:
oletko lukenut kansion readme-tiedostoa?

choldgraf:
The existence/location of that readme file is precisely the kind of thing that would be helpful in developer documentation ;-)

Issue #216, opened 2020-09-23
Nesting "margin" under h2 / h3 headers influence location of margin unexpectedly


consideRatio:
**Describe the bug**

The location of a margin differs depending if you are nested under a `##` header or a `###` header. This can be seen here.

![jupyter-book-sections-width](https://user-images.githubusercontent.com/3837114/94018075-752a4f80-fdb0-11ea-82eb-ab36d5d0732b.gif)

**To Reproduce**

Render the following in a markdown file with `jupyter-book build`

    ## my-h2-header
    my-h2-text
    
    ````{margin}
    ```{admonition} Aligned under h2
    admonition text under h2
    ```
    ````
    
    ### my-h3-header
    my-h3-text
    
    ````{margin}
    ```{admonition} Aligned under h3
    admonition text under h3
    ```
    ````

**Expected behavior**

To align the margin output independent if nested under a H2 or H3 header.

**Environment**

- Python Version: 3.7.6
- Package versions or output of `jupyter-book --version`:

  ```
  Jupyter Book: 0.8.1
  MyST-NB: 0.10.1
  Sphinx Book Theme: 0.0.36
  MyST-Parser: 0.12.9
  Jupyter-Cache: 0.4.1
  NbClient: 0.4.0
  ```

- Operating System: Ubuntu 20.04

consideRatio:
> is the issue that there is an extra left margin when it's nested under one header but not another?

@choldgraf yes that they don't align is the issue as I see it, and that the reason relate to if they are declared under one header or another could of course make sense to adjust for, but I don't think so at the moment at least.

consideRatio:
@choldgraf thanks this resolved the issue I experienced!

PR #218, opened 2020-09-23
🐛 FIX: fixing section width

choldgraf:
this sets a right padding only on the top-most sections, rather than on all sections, which I believe is causing the bug that @consideRatio reported. @consideRatio wanna double check that it looks OK to you now?

see https://sphinx-book-theme--218.org.readthedocs.build/en/218/reference/markdown_limits.html#margin-under-lower-level-shouldn-t-have-different-left-alignment for an example

closes https://github.com/executablebooks/sphinx-book-theme/issues/216


pradyunsg:
Can we disable the codecov's comments? The reports are available in the CI checks. 🥺

choldgraf:
sure, I just don't know how to do that but feel free to open a PR that does!

pradyunsg:
#220 does that! ^>^

PR #220, opened 2020-09-23
♻ IMPROVE: Disable codecov's comments

pradyunsg:


pradyunsg:
Based on https://docs.codecov.io/docs/pull-request-comments

choldgraf:
oops I picked the wrong word for the merge commit 😬 but thanks for the improvement!

pradyunsg:
It's an improvement. XD

chrisjsewell:
Erm what, why does this helpful information need to be disabled. I will most likely be reverting this

choldgraf:
I think @pradyunsg's reasoning is that it's redundant information with the tests that are in the status bar, and takes up a big amount of space in the PR comments etc. It seems like a reasonable change to me given that we are now testing for regressions in code coverage, I believe.

Is there a strong reason that you want to keep it? I merged because I didn't think this would be controversial, it was just the default for codecov, but we can discuss as a group if you have other preferences.

pradyunsg:
Indeed.

Relevant context:

- the initial comment: https://github.com/executablebooks/sphinx-book-theme/pull/218#issuecomment-697942984
- the codecov checks page, for that PR: https://github.com/executablebooks/sphinx-book-theme/runs/1156980593

chrisjsewell:
> is there a strong reason you want to keep it?

Is there a strong reason to remove it? Why should I now have to delve into the tests to find the coverage report?
It's the default for a reason, because it provides helpful information.

Why does it matter if there is an extra comment in PRs, when has that ever been an issue 🤷‍♂️

choldgraf:
I don't really care that much to be honest. I didn't realize that it was a comment that you often look at (I find myself generally skipping over it and only looking at the coverage if one of the cov tests fails, and I generally think it's good to only put information in front of eyeballs if it's genuinely useful or else it reduces signal-to-noise).

That said, if it's something you really feel strongly about then we can revert the PR. I didn't think it'd be a big deal.

chrisjsewell:
It's just a weird change to make; you should justify diverging from the defaults, not the opposite way round. 
Anyway, I'll leave it for now, but may revert at a later date

choldgraf:
I think we should have created an issue about it first, @pradyunsg asked about it in https://github.com/executablebooks/sphinx-book-theme/pull/218#issuecomment-697942984 but that was just a comment in another PR. Apologies that this was jarring (again I didn't imagine it would be controversial, I wouldn't have merged if I thought there'd be any objection).

chrisjsewell:
Ah no it clearly not the end of world lol, I just wanted to mention don;t be suprised if I change it back at some point 😉

PR #225, opened 2020-09-29
👌 IMPROVE: declare parallel read safe

rscohn2:
I am getting this warning:
```
WARNING: the sphinx_book_theme extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
```
The PR declares the extension is [parallel read safe](https://www.sphinx-doc.org/en/master/extdev/index.html#extension-metadata), eliminating the warning and allowing parallel read.

I am not sure if the extension is parallel read safe, but it seems likely.


choldgraf:
ah yes - I suspect that it is as well. @chrisjsewell I feel like you have more experience w/ parallelism in Sphinx, any reason to think we can't set this to True?

chrisjsewell:
no it shouldn't be an issue, although of course you can always look to add a test case for parallel builds

rscohn2:
I added a test. It is very simple, but did find a problem in another extension. I tried, but was unable to add add it to pytest. You have to pass parallel=2 to the [constructor of Sphinx](https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/application.py#L141), but I don't know how to pass it through.
As an alternative, I added it directly to the github ci config.

I have also tested on a 1500 page doc and it works fine.

The force push was to cleanup the commits addressing ci issues.

chrisjsewell:
>  but I don't know how to pass it through.

Yeh unfortunately it is not available via the test app:
https://github.com/sphinx-doc/sphinx/blob/d8c006f1c0e612d0dc595ae463b8e4c3ebee5ca4/sphinx/testing/util.py#L101

sphinx_book_theme/__init__.py
```diff
@@ -386,3 +386,5 @@ def setup(app: Sphinx):
     app.connect("html-page-context", add_to_context)
 
     app.add_directive("margin", Margin)
+
+    return {"parallel_read_safe": True}
```
chrisjsewell, line 390:
```suggestion
    return {"parallel_read_safe": True, "parallel_write_safe": True}
```

While we are at it. pydata-sphinx-theme (which this derives from) sets both of these also, so again I don't believe it should be an issue

chrisjsewell (changes requested):

.github/workflows/tests.yml
```diff
@@ -41,6 +41,9 @@ jobs:
     - name: Run pytest
       run: >
         pytest --durations=10 --cov=sphinx_book_theme --cov-report=xml --cov-report=term-missing
+    - name: Run parallel build test
```
choldgraf, line 44:
```suggestion
    # Running the parallel build test on its own because we cannot pass parallel config to Sphinx via pytest
    - name: Run parallel build test
```

does this correctly encapsulate why we're doing this outside of pytest?

rscohn2: I will make the comment more explicit about the need to pass parallel=2 to the sphinx constructor, including the [link](https://github.com/sphinx-doc/sphinx/blob/d8c006f1c0e612d0dc595ae463b8e4c3ebee5ca4/sphinx/testing/util.py#L101)
I also will move the test into pytest as I did for this other extension:
https://github.com/readthedocs/sphinx-notfound-page/blob/47887ecb35f42da29dd111f8a281a22ad4bf7412/tests/test_urls.py#L33

chrisjsewell: err, maybe. I specifically moved away from running sphinx using subprocess in pytest, because it is harder to debug and breaks in Windows ae76e1286d6378da296253dce0ce40305d8c5197

rscohn2: I moved it before I saw your reply. I can put it back to the yml if you prefer. I will also try to add support to SpinxTestApp, but it might take a long time before it shows up in a release.

chrisjsewell: Nah it's ok I just wanted to point this out

rscohn2: OK. The SphinxTestApp PR is https://github.com/rscohn2/spion/sphinx/pull/8257

choldgraf:
alright I think that we're all 👍 on the changes and tests are all happy, so I'm gonna merge this in. Thanks very much @rscohn2 !! much appreciated :-)

Issue #112, opened 2020-05-30
Collapsible ToC headings


akhmerov:
I propose to allow making some toc sections collapsible and collapsed by default.

While the default works well, collapsible toc sections would be potentially useful for appendices/exercise solutions (my personal use case)/bonus chapters.

Edit: [readme.com](https://docs.readme.com/metrics/docs) has a good example of collapsing entries in the section-pages-tree (left sidebar):

<img src="https://user-images.githubusercontent.com/2997570/89730335-9ff55a00-da35-11ea-938e-cf7246db0a4b.png" width="400px">

choldgraf:
Could you describe a bit more what you're imagining? I think it sounds like a good idea, I'm just not sure exactly what you mean

akhmerov:
e.g. like the left navbar [here](https://solidstate.quantumtinkerer.tudelft.nl/) (that's mkdocs-material)

choldgraf:
ah - and when you say "toc" do you mean the left or the right?

akhmerov:
Indeed, I mean the left one.

choldgraf:
gotcha - +1 from me, though this would require some reconfiguration of how the left sidebar works I think it could be useful

pradyunsg:
readme.com is another documentation-oriented 3-column page design, that allows for collapsing entries in the section-pages-tree (left sidebar).

Screenshow w/ link to page: https://gist.github.com/pradyunsg/eeb74ac5b580d502c3d0e248310f273f#readmecom

choldgraf:
thanks for the link!

also think that this would be easier now that we're using the Sphinx TOCtree object instead of a custom one 👍

pradyunsg:
Hurray! Sorry if I was unnecessary noise in the PR!

chrisjsewell:
> Hurray! Sorry if I was unnecessary noise in the PR!

You should be 😝

choldgraf:
I think what @chrisjsewell means to say is "thanks for providing your perspectives and opinions in these PRs, it is much appreciated even if not directly incorporated into the PR :-)"

chrisjsewell:
No I don't think so 😂

PR #226, opened 2020-10-05
✨ NEW: Collapsible lists in sidebars

AakashGfude:
A minimalistic collapsible TOC heading feature which looks like:

Screen Shot 2020-10-05 at 11 02 31 am
 

The transition speed of opening and closing the panel is in the "slow" setting and the speed can be increased. 

One abruptness in the whole experience is that since the clicking on a heading leads to the loading of the heading page, the panel does not open/close (due to obvious reasons). Which we can either leave as it is, or go beast mode and make this theme a single page [application](https://developer.mozilla.org/en-US/docs/Glossary/SPA) (probably not)

Tests have not been implemented yet.  Will do it once the design/functionality is finalized.

cc: @choldgraf  @akhmerov @pradyunsg

fixes https://github.com/executablebooks/sphinx-book-theme/issues/112


pradyunsg:
Ooooo, nice!

My only suggestion would be to make the "indicator" for collapse floated to the right -- like in [Just the Docs](https://pmarsceill.github.io/just-the-docs/).

AakashGfude:
Thanks, @pradyunsg for the suggestion. Have added the change. Lemme know what you think.

src/scss/_main.scss
```diff
@@ -1005,9 +1010,17 @@ nav.bd-links {
         > a {
             font-size: 0.9em;
         }
+        .fa-chevron-up, .fa-chevron-down {
+            float: right;
+            font-size: 1em;
+            margin-top: 0.4em;
+            margin-right: 0.4em;
```
pradyunsg, line 1017:
```suggestion
            font-size: 0.75em;
            margin-top: 0.5em;
            margin-right: 1.25em;
```

Before:

Screenshot 2020-10-05 at 7 48 37 PM

After:

Screenshot 2020-10-05 at 7 50 53 PM

You might even want to use 1.5em for the right margin, although the spacing when the scrollbar is not hidden/overlaid.

pradyunsg (commented):
I like the look and 2 more suggestions:

- The current "slow" speed is _too slow_, and I think a faster animation (or no animation) would be much better. The user is trying to navigate the tree, and is not here to watch animations. :P 
- The "chevron" feels too big and also gets hidden underneath the MacOS's hidden-unless-scrolling scrollbar. Inline suggestion to work around this below.

choldgraf:
This looks really nice! Thanks @AakashGfude for the work! In general I think this functionality is just what we want and I like the minimalist approach.

Agree with @pradyunsg's comments re: size and speed, I think some minor tweaks would improve things a bit

How do buttons like these get added to sections? e.g. I noticed that the sub-sub-sections don't have the button so just curious how the behavior should be controlled.

AakashGfude:
Thanks @pradyunsg for the detailes suggestions. will implement them. 

> * The current "slow" speed is _too slow_, and I think a faster animation (or no animation) would be much better. The user is trying to navigate the tree, and is not here to watch animations. :P

I was doing it for the kids. but I guess I had the user base a bit wrong. 😅 


> How do buttons like these get added to sections? e.g. I noticed that the sub-sub-sections don't have the button so just curious how the behavior should be controlled.

@choldgraf at present the buttons are also assigned to `li` tags wit `toctree-l1` as class. which is the first level of the lists. we can go deeper and assign it to `toctree-l2` etc as well if need be.

AakashGfude:
> You might even want to use 1.5em for the right margin, although the spacing when the scrollbar is not hidden/overlaid.

@pradyunsg  I probably would not go for too much margin values, because it varies so much between screens dimensions. on my laptop. the latest margin value already does a lot of separation 

Screen Shot 2020-10-06 at 10 54 26 am

chrisjsewell:
Looks good cheers @AakashGfude 
I'll play around properly with it at some point, but off the top of my head, have you checked how it degrades when JavaScript is disabled?

AakashGfude:
> I'll play around properly with it at some point, but off the top of my head, have you checked how it degrades when JavaScript is disabled?

@chrisjsewell the opening/closing of lists don't work. and the sidebar functions like before. the chevron icons are still there though.

chrisjsewell:
thanks for looking, yeh the main by is obviously that the site should still be useable if Javascript is disabled, but it sounds like that is not an issue

So, I've had a look now. A few issues:

1. The top-level section titles should also be collapsible, i.e. `Main Docs` and `Reference Items` in the documentation
2. There is no nested collapsible sections, like in the [just-the-docs](https://pmarsceill.github.io/just-the-docs/docs/ui-components) documents. The nesting level of collabsible headings could perhaps be a configuration variable (in theme.conf)
3. As you mentioned, currently the collapsible sections are also the URL links, so exposing a section means you have to navigate to it. That **has** to be fixed before merging.
   There are maybe two solutions to this; make only the chevron itself clickable to toggle the collapse, as is done in the just-the-docs example, or make collapsible titles not URL links, like is done in my gold standard for these collapsible ToCs; the VS Code documentation: https://code.visualstudio.com/docs

AakashGfude:
Aweome @chrisjsewell . great points.

> 1. The top-level section titles should also be collapsible, i.e. `Main Docs` and `Reference Items` in the documentation.

This sounds good.

> 2. There is no nested collapsible sections, like in the [just-the-docs](https://pmarsceill.github.io/just-the-docs/docs/ui-components) documents. The nesting level of collapsible headings could perhaps be a configuration variable (in theme.conf)

We should definitely handle all expandable sections. not sure if it should be in theme.conf though. As this extra config may not be that useful or used. Everyone would probably just want all of their sections to be collapsible?


>    There are maybe two solutions to this; make only the chevron itself clickable to toggle the collapse, as is done in the just-the-docs example, or make collapsible titles not URL links, like is done in my gold standard for these collapsible ToCs; the VS Code documentation: https://code.visualstudio.com/docs

From the looks of it doing what just-the-docs does atm, won't change anything in comparison to present implementation from the user experience point of view? i.e., it navigates to the section and opens it up as well.  
For VS code type implementation, we will then have to make an extra entry for the title URL as well in the section ?

chrisjsewell:
>  just-the-docs also navigates to the section and opens it up.

No it doesn't, not if you click directly on the chevron

chrisjsewell:
In just-the-docs, the chevrons are completely seperate (from the text), clickable SVGs, rather than fontawesome icons

AakashGfude:
ahh yeah. did not click directly in the chevron to open up a different section. my bad.

AakashGfude:
have made only the icons clickable like just-in-docs now. for everyone to try and give feedback on.

chrisjsewell:
The feedback is that it doesn't look lol. Your chevrons are still underneath the links

AakashGfude:
> The feedback is that it doesn't look lol. Your chevrons are still underneath the links

Oops haha. Sorry, in transit. Will reach a stable place and push the rearrangement of elements edit. Sorry for giving false hope.

AakashGfude:
@chrisjsewell should be in a working state now.

chrisjsewell:
Yep the chevron only collapsing works now thanks. I wonder if we should make that more intuitive though? Something like the cursor icon changes when you hover over it?

AakashGfude:
> Yep the chevron only collapsing works now thanks. I wonder if we should make that more intuitive though? Something like the cursor icon changes when you hover over it?

That's a good idea. some ways can be 1) increase the font size on hover 2) change color 3) change background-color

AakashGfude:
TODO:

- [ ] Add possible tests
- [ ] Make this PR mergeable

AakashGfude:
Have added color change when hovering on icons. Icon size increase and the border getting colored didn't look right. ideas on anything more dramatic are welcome.

src/scss/_main.scss
```diff
@@ -1011,6 +1017,18 @@ nav.bd-links {
...
         }
         font-size: 1em;
+        position: relative;
+    }
+
+    .fa-chevron-up, .fa-chevron-down {
+        position: absolute;
+        font-size: 0.75em;
+        right: 1rem; // gives the same padding as 1.25em
+        top: 0.5rem;
```
pradyunsg, line 1027:
```suggestion
        right: 0.5em;
        top: 0;
        padding: 0.5em;
```

Let's add a bit of padding and make the potential area you need to click bigger.

pradyunsg (commented):
If we're making the collapsing recursive, I think we should also start with the deeper levels collapsed by default.

chrisjsewell:
Looking better 😄 

I would definitely like to control the initial collapsed level though, i.e. I think for the documentation here, I think this is how the ToC should initialise:

image

chrisjsewell:
> I would definitely like to control the initial collapsed level though (via a config value)

Oh I just noticed this is similar to what @pradyunsg said, but yeh it should be configurable

choldgraf:
I think this looks really nice! Two quick thoughts:

- Agree about default collapse configuration. The pydata theme calls this `show_toc_depth` ([ref](https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/configuring.html#show-more-levels-of-the-in-page-toc-by-default)), maybe we can call it `show_navbar_depth`?
- What do folks think about *not* adding a collapse button to **part headers**? These headers will *always* be shown by default, so the button would only be used to *close* the section but not open it. I feel like this feature is more useful for folks that want to browse hidden sections without changing pages, and the extra button for each "part" header (which is unclickable anyway) adds some visual clutter

chrisjsewell:
> The contents of these headers will always be shown by default

Why should these alway be shown by default?
They're not in https://code.visualstudio.com/docs, which has "non-url" top-level headers.
In fact, like there, it would be ideal if these headers were actually "fully clickable", as opposed to only the chevrons for the "url" headers.

choldgraf:
I'm making the assumption:

- Those headers are not clickable (since Sphinx doesn't make them clickable, though I guess we can over-ride Sphinx)
- Those headers don't point to a specific page, they are just a word describing a collection of pages

I agree that in theory they could be clickable, we change how Sphinx handles the relationship between `:caption:` and toctree headers, but at least thus far the more common use-case has been to just have them as headers with shown links below, and I think we should be designing for the more common use-case at least initially. We can always add the ability to collapse those sections later on, or add a config value to make it possible.

chrisjsewell:
I don't mean clickable to link to a page, I mean clickable to un-collapse the content.
Again, I'd emphasise there are plenty of use cases where people may have enough of these part headers, such that they would want them all collapsed, as exampled in: https://code.visualstudio.com/docs

choldgraf:
Other ideas for how to de-clutter those buttons:

**Make the color slightly lighter so it doesn't pop as much**

```css
nav.bd-links .fa-chevron-up, nav.bd-links .fa-chevron-down {
    color: #ababab;
}
```

image

**Display only on hover**
```css
.collapsible-parent > i {
    visibility: hidden;
}

.collapsible-parent:hover > i {
    visibility: visible;
}
```

ORWMyT5mjN

chrisjsewell:
yep something like that sounds reasonable 👍

AakashGfude:
if we are controlling collapse/un-collapse in terms of levels now, what implications will it have to this config https://sphinx-book-theme.readthedocs.io/en/latest/layout.html#expand-sections-of-your-navbar. 

1) Do we want to keep it as is/remove it
2) Do welet users to specify subsections and top-level headers in it as well? (at present it takes just level 1 li)

pradyunsg:
TBH, I'm in favor of mimicking what [Furo](https://pradyunsg.me/furo/) and [Just the Docs](https://pmarsceill.github.io/just-the-docs/) do -- default to hidden unless the page you're on would be a part of that hierarchy.

It's a fairly sane default and avoids providing a knob to the user without any significant loss in functionality (which is a good thing IMO).

pradyunsg:
I'm also curious what the behavior is of this code when a given toctree does not have any captions.

choldgraf:
I'm +1 on deprecating the "expand sections" option as a part of this PR. I do not believe it is heavily used and this is a more elegant solution to a similar problem 👍 

So basically +1 to @pradyunsg's suggestion here: https://github.com/executablebooks/sphinx-book-theme/pull/226#issuecomment-706079541

AakashGfude:
Thanks, guys, for your inputs. 
Recent push has : 

-  all sections collapsed unless the current page is a part of that hierarchy or root

-  removed `expand_sections` config

@pradyunsg do you prefer keeping/removing the hover color in the `i` tag? since our texts have hover color, thought `i` tag should have as well. 

So, do we want a config variable as well to control the collapse? or it's good enough with this default behavior?

AakashGfude:
> I'm also curious what the behavior is of this code when a given toctree does not have any captions.

The chevrons are applied to and are clickable only for necessary sections of this one combined toctree.

pradyunsg:
/me feels like we're basically slowly making this the same as Furo, and it'd be nicer to re-use that entire chunk of code.

chrisjsewell:
there should definitely be a configuration for the initial depth of uncollapsed titles, since this will vary by book size.

BTW, I know you can click on the logo to get back to the index page, but it feels like it would be nice, at least for smaller books/sites, if that page could also be (optionally) added to the top of the ToC. It just feels a bit weird, when you only have a few pages, that one is missing from the ToC

chrisjsewell:
Ahhh, I missed that! Also in the documentation, we should mention about how to configure title, logo and favicon, i.e. link to the sphinx documentation on `html_title`, `html_logo` and `html_favicon`, since its a standard thing that people will want to do.

AakashGfude:
> * I'm gonna be annoying and contradict myself earlier :-)  I think we should actually show the carets even without hovering over sections - otherwise users will have to hover over sections themselves to know which ones are click-able, and we kinda lose the benefit of the dropdowns in the first place IMO. What do you think?

No worries at all. I was in favor of showing them permanently from start. but wanted to visually compare the experience difference. I think trying out different things visually and then settling with one is a good approach.
 
> * Agree the captions should be openable via a click on their title, not just the caret
will implement this one in a while.

have added a configuration option now : `show_navbar_depth` , with default value of 2.  Basically, when it's 0, all sections are collapsed. and as the number increases more levels are shown.  
If this feature looks good, then will add tests, docs.

chrisjsewell:
Thanks @AakashGfude, looks like you need to trigger a new RTD build though, since there's been a failure of a notebook execution (probably just an intermittent thing).

It would be good though to add this to the docs `conf.py`:

```python
execution_show_tb = "READTHEDOCS" in os.environ
```

This prints the failure traceback, so you can see it on the RTD build page, where obviously you don't have access to the `../reports/markdown_limits.log` it refers to.

choldgraf:
https://github.com/executablebooks/sphinx-book-theme/pull/226#issuecomment-708716740

oh shit that's a pretty cool pattern, I didn't know about that

chrisjsewell:
image

AakashGfude:
Thanks, @chrisjsewell, looks like the docs are building in the latest push. 

> It would be good though to add this to the docs `conf.py`:
> 
> ```python
> execution_show_tb = "READTHEDOCS" in os.environ
> ```

should I add this permanently to the docs with this PR?

chrisjsewell:
> should I add this permanently to the docs with this PR?

yep 👍

choldgraf:
One quick nit: the slight offset here is kinda driving me nuts could you make those aligned? 🙃  

image

AakashGfude:
The latest push:

1. makes the whole caption clickable to open/close
2. style adjustment as asked in https://github.com/executablebooks/sphinx-book-theme/pull/226#issuecomment-708838809, the reason behind this anomaly I could not find.

choldgraf:
I think it is looking really nice ✨ 

@AakashGfude one thing that's missing is documentation for the new `navbar_depth` feature...could you add that to the `configure` section?

Is there anything else left on the to-do list for this PR or do we want to punt some things to other issues? What do folks think?

chrisjsewell:
Yep tested it locally (with good old `tox -e docs-live`), changing the `show_navbar_depth` value, and it correctly updates the expansion 👍

chrisjsewell (approved):

choldgraf:
Agree - I am +1 when this is documented!

chrisjsewell:
oh yeh just pre-approving, assuming the docs will be updated

AakashGfude:
Have written documentation but not sure how clear it is. Someone with better writing skills, please check 😬 .

chrisjsewell:
Yep looks fine to me. Let's get this show on the road!

choldgraf:
One question - I noticed that the default depth is 2. Which means that by default we'll show top-level chapters **and** each of their nested sections (but we won't show sub-sections). That deviates a bit from the typical Sphinx theme behavior, which only shows top-level behavior, right?

My intuition is that we don't want the SBT to behave that different from most other Sphinx sites. What do you all think about setting the default level to `1`?

chrisjsewell:
Is this not also in some way controlled by the `:maxdepth:` option of the toctrees: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree? I guess that sets what is "available" in the fully expanded ToC

choldgraf:
Ah - I always thought that `:maxdepth:` only applied to the *inserted* TOCTree, not the one in any sidebars...

chrisjsewell:
Yeh not sure, what be never really thought about it TBH

AakashGfude:
`:maxdepth:` does not change anything about the `toctree` object we capture using bs4. which is what we use to render the sidebar.

AakashGfude:
> `:maxdepth:` does not change anything about the `toctree` object we capture using bs4. which is what we use to render the sidebar.

Also because we are explicitly specifying the `maxdepth` value as -1 while capturing it .

```
# Grab the raw toctree object and structure it so we can manipulate it
        toc_sphinx = context["toctree"](
            maxdepth=-1, collapse=False, titles_only=True, includehidden=True
        )
        toctree = bs(toc_sphinx, "html.parser")
```

chrisjsewell:
> :maxdepth: does not change anything about the toctree object we capture using bs4. which is what we use to render the sidebar.

I see, I guess the question is, should there ideally be any integrations with `:maxdepth:` (and the other hard-coded options)?
Is this taking away any of sphinx's functionality from the user?

AakashGfude:
> I see, I guess the question is, should there ideally be any integrations with `:maxdepth:` (and the other hard-coded options)?
> Is this taking away any of sphinx's functionality from the user?

I think we can utilize `maxdepth` value to set the cap on the levels to include in the sidebar. and then configure on that sidebar which levels should be expanded/collapsed using `show_sidebar_depth`?  
But probably sidebars are meant to include all levels?

AakashGfude:
Have made the default level as 1 now.

chrisjsewell:
> But probably sidebars are meant to include all levels?

For sure its more important when you don't have access to these collapsible ToCs

But, I guess for maximum configuration, it would be ideal to be able to say "under this section I want everything collapsed, and for this section I want to uncollapse one level, ...", i.e. it would be used like a per-toctree `show_sidebar_depth`.
But not saying that needs to be implemented in this PR

choldgraf:
> it would be ideal to be able to say "under this section I want everything collapsed, and for this section I want to uncollapse one level, ...", i.e. it would be used like a per-toctree show_sidebar_depth.

I'm +1 on this as well (assuming this is how Sphinx expects us to use `maxdepth`). I think it's best as a new issue so that we can use this nice new feature ✨

choldgraf:
ah - just noticed a problem. If I'm on a page then its children are not collapsed:

image

e.g.: https://sphinx-book-theme--226.org.readthedocs.build/en/226/reference/subchapter1b/index.html#

when I click that link, the sub-pages underneath do not show up

AakashGfude:
> when I click that link, the sub-pages underneath do not show up

ooh yeah, grrr, forgot about this feature. :(

AakashGfude:
@choldgraf, Have added the feature to expand immediate child ul as well on clicking a link, or page load. let me know how it feels.

choldgraf (approved):
I think it looks great!

choldgraf:
alright! Features in, tests happy, docs written, thumbs up given, let's merge it in and start playing around with it! 🎉 

Thanks @AakashGfude for the improvement!

chrisjsewell:
cheers!

choldgraf:
I kinda wanna cut a little release...@chrisjsewell did the sidebars update give you any trouble? I wanted to make sure we didn't introduce any bugs when that one got merged.

chrisjsewell:
https://aiida-blog.readthedocs.io/en/chris-branch/stories/index.html

choldgraf:
they look pretty good to me!

PR #228, opened 2020-10-06
👌 IMPROVE: Add missing aria-label to fullscreen button

foster999:
Google Lighthouse highlighted the lack of aria-label as an accessibility issue.
This PR adds an accessible label for the "Fullscreen mode" button.


choldgraf:
nice - many thanks for noticing this!

foster999:
No worries 😊

Issue #244, opened 2020-11-01
Disable 'Download this page...' dropdown menu


bknaepen:
**Is your feature request related to a problem? Please describe.**

I would like to be able to disable the download dropdown menu from the topbar.

**Describe the solution you'd like**

An extra option in the _config_yml similar to:

```
use_issues_button: false
```

**Describe alternatives you've considered**

Currently I have added this to my css file to hide the menu:

```
[aria-label="Download this page"] {
display: none;
}
```

Cheers,
Bernard.

choldgraf:
closed by #245 !

PR #245, opened 2020-11-01
✨ NEW: Option for download button

bknaepen:
This relates to the following open issue:

https://github.com/executablebooks/sphinx-book-theme/issues/244

I am completely new to this package but this proposal seems to be functional. I just duplicated lines I found for the issue button. 

The tests are failing but I was unable to fix this.


src/jinja/theme.conf.j2
```diff
@@ -17,6 +17,7 @@ navbar_footer_text =
 extra_navbar = Theme by the <a href="https://ebp.jupyterbook.org">Executable Book Project</a>
 extra_footer =
 use_issues_button = False
+use_download_button = False
```
choldgraf, line 20: let's set this to `True` by default

bknaepen: That did it :-)

sphinx_book_theme/theme.conf
```diff
@@ -17,6 +17,7 @@ navbar_footer_text =
 extra_navbar = Theme by the <a href="https://ebp.jupyterbook.org">Executable Book Project</a>
 extra_footer =
 use_issues_button = False
-use_repository_button = False
+use_download_button = False
```
choldgraf, line 20: hmm - `use_repository_button` should remain `False` (the default that we currently have), and `use_download_button` should be `True` (also our current default behavior)

Can you also add documentation for this feature to https://github.com/executablebooks/sphinx-book-theme/blob/master/docs/configure.md ?

bknaepen: Sorry, I obviously edited the wrong line 🤦🏻‍♂️. I hope I got it right this time...

I'll proceed to the documentation and be back.

pradyunsg:
Another `The message was: Cell execution timed out.`. Closing-reopening to re-trigger CI.

choldgraf:
Thanks very much for this! RTD timeout errors fixed so let's get it in!

bknaepen:
@choldgraf I'm glad this feature was found useful, thanks for helping out :-). You mentioned above that a validation test would also be useful. I could work on this in the coming days when time permits but I need to dig a bit more on how to do this. What do you think?

choldgraf:
that would be great @bknaepen . I'd recommend checking out how the tests work for similar features: https://github.com/executablebooks/sphinx-book-theme/blob/master/tests/test_build.py#L217 and try to recreate those patterns. Wanna give that a shot?

bknaepen:
@choldgraf Yes I will try ;-)

PR #246, opened 2020-11-01
👌 IMPROVE: default colors for admonitions

choldgraf:
pydata sphinx theme recently made colors configurable via CSS variables, and also changed the default colors. This just keeps our default admonition colors the same as before the pydata switch


choldgraf:
gonna merge this now since our colors are wrong currently on `master`!

PR #249, opened 2020-11-04
🐛 FIX: Fixing colors for future theme update

choldgraf:
An attempt at fixing colors since the pydata theme changed the default colors but also made it possible to configure them.

This shouldn't actually affect the docs until the next version of the pydata theme is released, but shouldn't break anything in the meantime


choldgraf:
merging this one in - I think there might be another "timeout" bug in one of our documentation pages but it's not related to this theme so we can fix in a subsequent PR if it pops up

PR #259, opened 2020-11-11
👌 NEW: Added tag for cell-input

AakashGfude:
As suggested by @mmcky , we should also enable input tags to have a fixed height and scrollable, using tags.

With this PR,

```
```{code-cell} python3
---
tags: [scroll-input]
---
```

will work the same for input cells, as scroll-output was working for output cells.

If this sounds like a good addition, then we can go ahead and add it to jb docs like in this PR: https://github.com/executablebooks/jupyter-book/pull/1075/files. (although I think that these tags should be in sphinx-book-theme docs)


choldgraf:
+1 from me! Wanna add some docs here and then we can tackle a PR in JB?

AakashGfude:
@choldgraf sorry did not get you? Add docs in `sphinx-book-theme` as well?

choldgraf:
Yep we should document in both places, since people may use this theme independent of jupyter book

AakashGfude:
@choldgraf same goes for `scroll-output` then as well? Should I add docs for both here?

choldgraf:
Yeah as a general rule we document everything in both spots. Longer term we could think of ways to streamline the docs but for now I think we should make sure the information is in both spots

AakashGfude:
@choldgraf have finally added the docs 😅 . What do you think?

choldgraf:
sorry for the slow turnaround on this - but this looks great @AakashGfude , thanks for the improvement 🎉

PR #260, opened 2020-11-11
Doc download

bknaepen:
The content of the whole topbar is compared when setting use_download_button on or off.

Tests re inspired from a similar tests for the edit issue button.


choldgraf:
very nice! this looks great - one thought, can you collapse them into a single `test_topbar_download_button` test, rather than having one for "on" and one for "off"? Just add a comment before each block to describe what you're testing for

bknaepen:
I tried this:

```
def test_topbar_download_button(sphinx_build_factory, file_regression):
    """Test building with download button on or off."""
    
    # Button on
    confoverrides = {
        "html_theme_options.use_download_button": True,
    }
    sphinx_build = sphinx_build_factory("base", confoverrides=confoverrides).build(
        assert_pass=True
    )

    source_btns = sphinx_build.html_tree("section1", "ntbk.html").find_all(
        "div", attrs={"class": "topbar-main"}
    )[0]
    file_regression.check(source_btns.prettify(), extension="_on.html", encoding="utf8")

    # Button off
    confoverrides = {
        "html_theme_options.use_download_button": True,
    }
    sphinx_build = sphinx_build_factory("base", confoverrides=confoverrides).build(
        assert_pass=True
    )

    source_btns = sphinx_build.html_tree("section1", "ntbk.html").find_all(
        "div", attrs={"class": "topbar-main"}
    )[0]
    file_regression.check(source_btns.prettify(), extension="_off.html", encoding="utf8")
```

But I then get an error message during the copytree operation for the second test. The folder test_topbar_download_button0 already exists. Is there a way to create a different test folder for the second test?

choldgraf:
ah hmmm - how about instead we *only* test for when the topbar is set to "off", since it is "on" by default and we are implicitly testing this in the default book build tests?

bknaepen:
@choldgraf The tests now only check the option to set the download button to off...

choldgraf:
Great - thanks very much!! 🎉

PR #268, opened 2020-12-05
🐛 FIX: fixing right toc whitespace overlap

### LINKED ISSUE #267: Table of Contents overlaps with sidebar content
I just noticed that our table of contents now overlaps with the sidebar even when it is closed:

image

could this be related to the recent topbar rework that @pradyunsg did?

choldgraf:
This is an attempt at fixing the right TOC, which currently has a white background that blocks any right margin content.

This PR:

- Makes the box shadow now span the whole topbar. This is because when the TOC *wasn't* shown, the shadow would "end" just before the TOC block instead of extending to the side of the page
- Makes the left sidebar have a white background (so that it blocks the shadow on the left)

However, there is one snag which is that currently the right TOC doesn't _quite_ cover up the topbar shadow:

image

@pradyunsg do you have any ideas? It would be great to get some help on this fix. You mentioned when you submitted the PR that you had this working, perhaps you remember what you had done before? Or maybe @AakashGfude has ideas? This is a pretty major bug we should fix because nobody's sidebar content will work properly

To see an example of this bug, see: https://sphinx-book-theme.readthedocs.io/en/latest/layout.html#content-sidebars

closes #267


pradyunsg (commented):
Looks A-OK to me?

choldgraf:
I think it looks fine except the drop shadow sticks out to the right of the top (see the image I posted). Any thoughts on how to fix this?

choldgraf:
@pradyunsg I think I might have fixed it in the latest commit, how does it look? There is a slightly hacky approach, which is to position the TOC at `-1em` in order to overlap the `1em` right padding that the body div has, and this makes the right TOC align with the right side of the page

pradyunsg:
I like it. :shipit:

choldgraf:
OK I'm gonna merge this in because it's fixing a kinda gnarly bug right now, and the new CSS rules seem to be behaving as we'd expect!

PR #279, opened 2021-01-21
📚 DOCS: Add custom css info to docs

### LINKED ISSUE #183: Make CSS easier to customize
There are some common customization points that we might predict for this theme. In particular, things like background color, fonts/font-size, etc.

Other than specifically adding configuration for this, another thing we can do is choose reasonable CSS rules and provide documentation to easily override them.

An example where we are sub-optimal is in the background color of the page. If somebody changes the background color of the content, they get results like this:

image

Because we're manually setting the background of some other elements. We should figure out a way to over-ride this naturally (I wonder if there's some way we could let people over-ride with CSS variables?)

rabernat:
I found it very easy to customize my CSS based on the instructions in https://github.com/executablebooks/sphinx-book-theme/issues/183#issuecomment-764672888.

This documentation update therefore potentially resolves #183.

### DIFF
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -225,7 +225,7 @@ Value of 0 will collapse all the sections and sub-sections by default.
 
 ## Customize CSS
 
-To customize the look of your site further, you can cusomize your CSS stylesheets,
+To customize the look of your site further, you can customize your CSS stylesheets,
 as described in the [ReadTheDocs Docs](https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html#adding-custom-css-or-javascript-to-sphinx-documentation).
 
 First, create a CSS file and place it in `_static/custom.css`.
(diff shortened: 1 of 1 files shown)


docs/configure.md
```diff
@@ -222,3 +222,23 @@ show_navbar_depth = <level>
 
 The default value is 1, which is used in this documentation.
 Value of 0 will collapse all the sections and sub-sections by default.
+
+## Customize CSS
+
+To customize the look of your site further, you can cusomize your CSS stylesheets,
```
choldgraf, line 228:
```suggestion
To customize the look of your site further, you can customize your CSS stylesheets,
```

choldgraf:
looks great - thanks for this improvement @rabernat !! 

It is always tricky to know how much information about "core sphinx config" to include in theme-specific docs, but I agree that customizing CSS is a big one and worth including.

PR #280, opened 2021-01-28
option to remove full screen button

lberndsenbaken:
I think this should provide an option to remove the full screen button in the top bar


choldgraf:
Ah I can't believe that I missed this PR - thanks for opening it! This was way out of date with master, so I rebased and cleaned up the errors, and merged it in here: https://github.com/executablebooks/sphinx-book-theme/pull/328

thanks very much for the contribution!

PR #283, opened 2021-02-09
Update translation/french: suggest edit, open issue

Naereen:
Hello there,
A tiny contribution to this sphinx-book-theme:

- The current translation was the same as "open issue" -> "problème ouvert" (which is also wrong, see next commit),
- The current translation of "open issue" -> "problème ouvert" was wrong (it interprets "open" as an adjective and not a verb).

Thanks in advance, regards! -- @Naereen


chrisjsewell:
thanks @Naereen! Looks like the "suggest edit" was accidentally a copy of "open issue" 🤦 . I've opened #284 and added your suggestions to it

PR #284, opened 2021-02-09
🐛 FIX: translation of suggest edit

chrisjsewell:
Also fix french translations (thanks to @Naereen)


Naereen:
Thanks @chrisjsewell!

PR #287, opened 2021-02-13
FIX: Hide TOC for "tall" full-width elements

shvenkat:
I have a full-width image that is also long in the vertical dimension. As I scrolled down, the image would come into view from below, and when the top margin overlapped the "TOC zone", the TOC would be hidden. That's pretty nifty, thanks! But as the top margin disappeared above, the TOC was displayed over the middle of the image. Then as the bottom of the image overlapped the "TOC zone", the TOC would be hidden again. This commit fixes the `removeToc` boolean to include the "tall" full-width element case.


shvenkat:
Regarding the failing pre-commit check above ... I'm guessing there's a pre-commit hook to minify the JS. I made this commit on this site, which is why the hook didn't run. I don't have `pre-commit` set up, and I wanted to quickly commit the fix, before getting back to my project. If someone could please run the hook, that would be great. Thanks!

choldgraf:
@shvenkat thanks for the fix - I wonder if you could provide an example of an image that shows the unintended behavior you describe at first, so that we can include it in the docs and then use it as an "eye-test" to make sure that the TOC behavior is, in fact, fixed. Does that make sense?

shvenkat:
Here are two examples. The dimensions are 1000x1500 pixels (WxH) and 1000x500. On my screen, both images are too wide to fit in the central text column. Imagine a long page with text and tables at the top and these images below. With the "full-width" CSS class, both will cause the TOC to hide as you scroll up and these images come into view. But the 1000x1500 one is so tall that at some point it spans the height of the screen, filling up the view. This is when the TOC un-hides itself, and this is the issue this PR fixes. The 1000x500 one never spans the height of the screen and never fills up the view, and so has no issue.

I hope these examples help. Let me know if this doesn't make sense.

wide_and_tall
wide

choldgraf:
This looks good to me - I'd like to make a few tweaks to the javascript here but will do that in a follow-up PR!

choldgraf (approved):
thanks for the fix!

PR #291, opened 2021-02-18
🔧 MAINT: Loosen importlib-resources pin

bollwyvl:
This is another solveability experiment, from https://github.com/conda-forge/sphinx-book-theme-feedstock/pull/5.

See also:
- https://github.com/executablebooks/jupyter-book/pull/1221

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(
         "click",
         "pydata-sphinx-theme~=0.4.1",
         "beautifulsoup4>=4.6.1,<5",
-        'importlib-resources~=3.0.0; python_version < "3.7"',
+        'importlib-resources>=3,<4; python_version < "3.7"',
     ],
     extras_require={
         "code_style": ["pre-commit~=2.7.0"],
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(
         "click",
         "pydata-sphinx-theme~=0.4.1",
         "beautifulsoup4>=4.6.1,<5",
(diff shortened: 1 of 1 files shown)


choldgraf:
@bollwyvl can you give some context for this?

bollwyvl:
Since a lot of folk will expect to be able to install downstreams of this, e.g. `juptyer-book` _next to their code_, these kinds of tight pins will introduce all kinds of extremely mysterious issues.

A pin of `~=3.0.0` (translating to `>=3.0.0,<3.1`) means _exactly one version, ever_ can be installed, as it's _highly_ unlikely the upstream will publish `3.0.x` (they're already on `3.7.0`). It's unlikely there will be bug fixen, etc.

chrisjsewell:
Yeh fair, this isn’t problem to merge. It’s only anyway required for python 3.6 (part of core after that) and python 3.6 will be end-of-life at the end of this year

choldgraf:
I'm not sure I quite understand the implications but trust both of you two than I do myself haha. The error is the sphinxcontrib-bibtex bug, not related to this, so I'm gonna merge!

@chrisjsewell is this a change that we should implement throughout the EBP repos?

chrisjsewell:
> is this a change that we should implement throughout the EBP repos?

well as mentioned, it is an issue that will "fix itself" as people move to using python >3.6, so its not really a priority

choldgraf:
ah I see so once we drop support for 3.6 then this will no longer be an issue at all

Issue #145, opened 2020-08-06
Design: (right sidebar) Change "On this page" to "Contents" (or "Table of Contents")


pradyunsg:
**Describe the changes you'd like**

Title says it all.

**Describe alternatives you've considered**

None. Status quo is OK. Suggested change is nicer, but I'm *obviously* biased. :)

**Additional context**

This seems to be the phrasing for the in-page-navigation-tree on all 3-column documentation-oriented themes that I know of (except this theme and the pydata theme). Sample Screenshots: https://gist.github.com/pradyunsg/eeb74ac5b580d502c3d0e248310f273f

I'm happy to file a PR for this, if there's interest in this idea. :)

choldgraf:
I think it's a reasonable idea. My question is what those themes call the left sidebar, in that case. I just wanna make sure that our naming is consistent so we don't confuse the mental model that folks have.

pradyunsg:
Navigation is what I've seen most often (the default Sphinx theme, and the entire lineage of themes that led to it, uses that).

choldgraf:
Cool - then we should just make sure it's consistent in our docs etc, and I'm +1 on making it "contents" or "table of contents", whichever you think makes most sense.

chrisjsewell:
not table of contents (that confuses it with the left side bar)

pradyunsg:
"Contents" it is. :)

chrisjsewell:
yeh or "Page Contents", no strong opinion though

jorisvandenbossche:
BTW, feel free to also propose such changes in the pydata-sphinx-theme!

PR #293, opened 2021-02-27
replace 'Contents' with 'On this page'

### LINKED ISSUE #290: Single content option
I am using this theme via jupyterbook here: https://vita-epfl.github.io/openpifpaf/dev/intro.html

One feedback that I got is that the two navigation bars on the left and right are confusing and whether it would be possible to just have a single one.

Alternatively, it would be nice to change or customize the title of the right navigation bar: "Contents" seems more appropriate for the left navigation bar. I would like to name the right one "On this page".

svenkreiss:
Fixes #290.

### DIFF
--- a/sphinx_book_theme/__init__.py
+++ b/sphinx_book_theme/__init__.py
@@ -252,7 +252,7 @@ def add_to_context(app, pagename, templatename, context, doctree):
 
         out = f"""
         <div class="tocsection onthispage pt-5 pb-3">
-            <i class="fas fa-list"></i> { context["translate"]('Contents') }
+            <i class="fas fa-list"></i> { context["translate"]('On this page') }
         </div>
         <nav id="bd-toc-nav">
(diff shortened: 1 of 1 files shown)


choldgraf:
Hmmm I think that rather than changing this (which would also change it for everybody else), I’d recommend making this configurable. We could add a config value to the theme options that would override the default title for the the within page toc. Then you could override as you wish. Want to give that a shot?

chrisjsewell:
I would point out that we literally already changed this in the opposite direction: https://github.com/executablebooks/sphinx-book-theme/issues/145

svenkreiss:
I wasn't aware of #145. Obviously I disagree with the outcome of that discussion :)

choldgraf:
OK - I opened up #297 to track making this configurable, so going to close this PR. Thanks @svenkreiss for bringing it up!

Issue #297, opened 2021-03-01
Make the in-page TOC title configurable


choldgraf:
Here we hard-code the title of the in-page TOC:

https://github.com/executablebooks/sphinx-book-theme/blob/c9fae28a006ddc7ba79b53a7b50eff3f011ff5cd/sphinx_book_theme/__init__.py#L255

If this were configurable, e.g. via:

```
html_theme_options = {
	"toc_title": "Some other title",
}
```

Then users could change this as they wish.

PR #299, opened 2021-03-07
👌 IMPROVE: Making in-page TOC title configurable

AakashGfude:
This PR makes in-page TOC title configurable with the introduction of the following key:
```
html_theme_options = {
	"toc_title": "Some other title",
}
```

The default value of the title has been kept as `Contents`.

It will throw an error if the title is set to empty (I reckon a title is always necessary?)

@choldgraf is there any exception handling format we need to follow for this repo. or the way I have handled it is ok?

fixes https://github.com/executablebooks/sphinx-book-theme/issues/297

### DIFF
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -152,6 +152,18 @@ html_sidebars = {
 
 See the [Sphinx HTML sidebars documentation](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_sidebars) for more information.
 
+## Rename the right sidebar title
+
+You can rename the title of the in-page table of contents, in the right sidebar:
+
+```python
+html_theme_options = {
+    "toc_title": "{your-title}"
+}
+```
+
+The deafult value of the title is `Contents`.
+
 ### Default sidebar elements
 
 By default, this theme comes with these three theme-specific sidebar elements enabled on all pages:
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -127,6 +127,17 @@ html_theme_options = {
 }
 ```
 
+## Control the right sidebar items
+
+You can rename the title of the in-page table of contents, in the right sidebar:
+
+```python
+html_theme_options = {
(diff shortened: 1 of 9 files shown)


sphinx_book_theme/__init__.py
```diff
@@ -250,9 +250,14 @@ def add_header_level_recursive(ul, level):
...
 
+        if not context["theme_toc_title"]:
+            raise ValueError(
+                "'toc_title' key cannot be empty. Please set a non-empty value."
+            )
+
         out = f"""
         <div class="tocsection onthispage pt-5 pb-3">
-            <i class="fas fa-list"></i> { context["translate"]('Contents') }
+            <i class="fas fa-list"></i> { context["theme_toc_title"] }
```
choldgraf, line 260: Could we keep the `translate` bit? I think that this would attempt to auto-translate the title for versions of the book that are in different languages. Is that right? Is there any downside to keeping the `translate`?

AakashGfude: You are right, `translate` is a function with signature `<function get_translation.<locals>.gettext`. So, it should auto-translate.

AakashGfude: Have re-added it https://github.com/executablebooks/sphinx-book-theme/pull/299/files#diff-1463927e2c28d59a6cedd041803bdfd110650e761a5eda7b5db2b2d946b82949R261

docs/configure.md
```diff
@@ -127,6 +127,17 @@ html_theme_options = {
...
+
+You can rename the title of the in-page table of contents, in the right sidebar:
+
+```python
+html_theme_options = {
+    "toc_title": "{your-title}"
+}
+```
+
+The deafult value of the title is `Contents`.
```
pradyunsg, line 140:
```suggestion
The default value of the title is `Contents`.
```

choldgraf:
thanks very much!

PR #304, opened 2021-03-18
👌 IMPROVE: Master doc title specified in `_toc.yml` is visible in left-sidebar

AakashGfude:
As outlined in this issue https://github.com/executablebooks/jupyter-book/issues/1012 ,  master_doc title specified with `title` key in `_toc.yml` was not reflecting in the left sidebar. This PR fixes that.

Example:
```
- file: intro
  title: This is the intro
   ....
   ....
```

results in:-

Screen Shot 2021-03-18 at 2 40 00 pm

@choldgraf, the test infrastructure in this repo uses sphinx-build as it is `jb` agnostic. So, I won't be able to test this feature( which is directed only towards jb projects) here right? 

fixes https://github.com/executablebooks/jupyter-book/issues/1012

### DIFF
--- a/sphinx_book_theme/__init__.py
+++ b/sphinx_book_theme/__init__.py
@@ -125,6 +125,7 @@ def add_to_context(app, pagename, templatename, context, doctree):
             master_doc = config["master_doc"]
             master_doctree = app.env.get_doctree(master_doc)
             master_url = context["pathto"](master_doc)
+            # check for title in `_toc.yml` else get title from doctree
             master_title = app.config["globaltoc"].get("title") or list(
                 master_doctree.traverse(nodes.title)
             )
--- a/sphinx_book_theme/__init__.py
+++ b/sphinx_book_theme/__init__.py
@@ -125,8 +125,8 @@ def add_to_context(app, pagename, templatename, context, doctree):
             master_doc = config["master_doc"]
             master_doctree = app.env.get_doctree(master_doc)
             master_url = context["pathto"](master_doc)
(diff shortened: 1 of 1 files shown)


AakashGfude:
Yup. I think `globaltoc` won't work here.

sphinx_book_theme/__init__.py
```diff
@@ -125,10 +126,18 @@ def generate_nav_html(
...
             master_url = context["pathto"](master_doc)
-            master_title = list(master_doctree.traverse(nodes.title))
-            if len(master_title) == 0:
-                raise ValueError(f"Landing page missing a title: {master_doc}")
-            master_title = master_title[0].astext()
+
+            # check title in globaltoc (jb case) else get title from doctree
+            master_title = None
+            with suppress(Exception):
+                master_title = app.config["globaltoc"].get("title")
```
chrisjsewell, line 133: note this won't work quite the same with https://github.com/executablebooks/jupyter-book/pull/1293. But I would be interested if there is a less theme specific solution that could go straight into sphinx-external-toc 🤔

AakashGfude: @chrisjsewell you mean creating a global variable in sphinx-external-toc which has this value? and can be accessed directly in any theme.

choldgraf:
It seems like this one is a bit stale - and maybe addressed at the sphinx-external-toc level? I'll close it so that we have less cruft to worry about in a refactor, but we can re-visit if this issue still persists

PR #311, opened 2021-03-25
🐛 FIX: Fixing sidebar scroll

choldgraf:
closes https://github.com/executablebooks/jupyter-book/issues/1245


src/scss/_righttoc.scss
```diff
@@ -4,7 +4,7 @@
 
 .bd-toc {
     padding: 0px !important;
-    overflow-y: visible;
+    overflow-y: auto;
```
pradyunsg, line 7: auto should be the default, so omitting this rule might also work. :)

choldgraf: I kinda like it from an "explicit is better than implicit" standpoint...but could remove if you think it's not necessary

pradyunsg: I'm fine either way! :)

PR #314, opened 2021-03-26
📚 DOCS: Update extra_navbar documentation

fmaussion:
I may be mistaken but for me setting the variable in `conf.py` has no effect, while this does.


choldgraf:
excellent - thanks for catching this!

PR #317, opened 2021-03-28
Using a full-width topbar

choldgraf:
This is a follow up to https://github.com/executablebooks/meta/discussions/296, it is a minimal PR that implements a full-width topbar. There is plenty that we should do to improve the details here, but just wanted to get this up for feedback.

# Todo

- [ ] Re-work the topbar sections to use the latest PyData Sphinx Theme templating code
- [ ] Decide whether we like the topbar to span the whole width
- [ ] Depending on that decision, either revert it or not, either way - keep the topbar refactoring!


choldgraf:
closing this one in favor of #343

PR #324, opened 2021-04-09
First pass at upgrading pst

choldgraf:
This is a first pass at upgrading to the latest pydata sphinx theme, and removing our custom sidebar collapsing code to use it instead. This should remove a bunch of our CSS + `__init__.py` code, as well as some testing infrastructure for the sidebar collapsing logic.


src/scss/_leftbar.scss
```diff
@@ -136,71 +105,8 @@ div.navbar-brand-box {
 }
 
 nav.bd-links {
-    margin-left: 0px;
```
jorisvandenbossche, line 139: You will need to keep this line I think. Apparently in PST we have `margin-left: -15px;`, which makes that in the preview (https://sphinx-book-theme--324.org.readthedocs.build/en/324/) the sidebar navigation is shifted to the left

choldgraf:
OK updated the tests and cleaned up some of the code. I think that this is ready to go whenever we'd like to merge.

@DrDrij let me know if you'd like me to hold off merging so that this doesn't mess with some work you're doing.

Also speak now @executablebooks/ebpteam if you really wanted to be able to collapse the sidebar TOC entirely (so that only captions are shown and no links) via `navigation_depth = 0` since that is being removed here.

src/scss/_leftbar.scss
```diff
@@ -12,17 +12,6 @@
     transition: margin-left .2s ease 0s, opacity .2s ease 0s, visibility .2s ease 0s;
     z-index: 2000 !important;
 
-    &.collapse {
```
DrDrij, line 15: @choldgraf These styles add the slide animation to left sidebar.

PR #339, opened 2021-05-30
Fixing path to '_sources' directory when using builders besides 'html'

### LINKED ISSUE #338: dirhtml builder doesn't copy executed ipynb to sources, resulting in broken source download link
**Describe the bug**

I'm not sure if this is the right repo for this report, but when using the `dirhtml` builder with this theme and including `myst-nb` notebooks, I'm finding that the `Download .ipynb` link is broken because the executed `.ipynb` file is not copied over to the `_sources` directory. The `Download .md` link works fine and switching to the `html` builder also fixes the issue. If you have a sense of where this might be coming from, I'd be happy to look into the fix. Thanks!!

**Environment**

- Python Version: 3.9
- Package versions: sphinx-book-theme==0.1.0, myst-nb==0.11.1, sphinx==3.5.4
- Operating System: macos

dfm:
Moments after opening #338, I found the fix. When copying the `.ipynb` files, the destination path was hard coded as `html`, but should depend on the `builder` setting in `context`. I've tested that this fixes my use case, but let me know if it would be good to explicitly include a test!

Closes #338

### DIFF
--- a/sphinx_book_theme/launch.py
+++ b/sphinx_book_theme/launch.py
@@ -47,7 +47,7 @@ def add_hub_urls(
         # Figure out the folders we want
         build_dir = Path(app.outdir).parent
         ntbk_dir = build_dir.joinpath("jupyter_execute")
-        sources_dir = build_dir.joinpath("html", "_sources")
+        sources_dir = build_dir.joinpath(context.get("builder", "html"), "_sources")
         # Paths to old and new notebooks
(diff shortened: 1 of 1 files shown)


choldgraf:
Ah beautiful, thanks for catching this one. I think that reworking the tests for a non html builder would actually require a decent amount of reworking our test infrastructure, so I won’t ask you to do that 🙂 thanks for this fix!

PR #340, opened 2021-05-30
👌 IMPROVE: Updating sidebar collapsing behavior

choldgraf:
This makes two updates to the main-content behavior when the sidebar is collapsed:

- On narrow screens, it makes the main content more transparent, so that it more easily fades into the background and doesn't distract
- On wide screens, it gives the main content more left padding when there is *no sidebar* so that it's not right up at the edge of the screen


damianavila:
> On wide screens, it gives the main content more left padding when there is no sidebar so that it's not right up at the edge of the screen

I find the left padding a little bit weird... maybe my eyes are not expecting the main text to go right at the same time the TOC goes left... just my totally subjective feedback 😉  .

choldgraf:
@damianavila yeah I agree with you I think. Do you have thoughts on how we could improve the UX here? I feel like our current behavior makes the content go too-far to the left side of the page, and would benefit from being a bit more centered.

damianavila:
> @damianavila yeah I agree with you I think. Do you have thoughts on how we could improve the UX here?

Maybe some additional padding at the left so you don't go too far left? 
Horizontal space is less expensive than vertical one, IMHO.
So taking a little bit more left space in the default view should be OK?

choldgraf:
I wonder if we could instead use a combination of `margin-left: auto` and `max-width` on the main content div to get the same effect. For example, the [web accessibility initiative](https://www.w3.org/WAI/tutorials/page-structure/styling/#line-length) suggests that we should have a maximum of 80 characters or less on each line. Perhaps we could choose a maximum width of just above this, and cap the content div's width at that?

damianavila:
> I wonder if we could instead use a combination of margin-left: auto and max-width on the main content div to get the same effect. 

What do you mean with the "same effect"... because the thing I am criticizing is the "effect", regardless of how it is implemented :wink: .

choldgraf:
In this case, by "effect" I mean "more white-space on the left margin", but more generally I think this means "more well-structured layout when the sidebar is collapsed". I feel like collapsing the sidebar opens up horizontal space, but the content often opens to a width that isn't reader-friendly (e.g., most accessibility sites recommend keeping the width of content to around 80 characters) so I thought a maximum width + centering the content could be a way to make this behave more sensibly.

Maybe this is something we should discuss in an issue w/ mock-ups etc? If so I can remove the left-padding stuff from this PR and keep it just to the "transparency in the background" topic.

damianavila:
> Maybe this is something we should discuss in an issue w/ mock-ups etc? If so I can remove the left-padding stuff from this PR and keep it just to the "transparency in the background when sidebar is open" change.

I think that is a sensible idea. 
Btw, thanks for the additional context you posted above.

PR #341, opened 2021-05-30
📚 DOCS: Updating logos

choldgraf:
Updates the logos for our newly-designed style.


damianavila:
LGTM, but the branch seems to be outdated?

choldgraf:
updated the branch - there was one CI job failing but it was just because codecov wasn't responding properly, so I think it's OK

PR #343, opened 2021-06-09
Updating topbar to be full width

choldgraf:
This is another test to see how the topbar looks at full width.

Here's a GIF:

6P9pXpHhz0

Compare that to current topbar:

image


damianavila:
Mmm... I am not sure about this change... the left arrow has a "meaning" in the existing implementation and you are somehow losing that with this PR.

choldgraf:
@damianavila ya, I kind of agree...I think we could add more things in the upper left and that might make it feel less awkward up there. But just to be clear: this PR isn't a "planned merge" it is just playing around with UX/UI

damianavila:
> But just to be clear: this PR isn't a "planned merge" it is just playing around with UX/UI

Sure, my feedback was in that playing context...

choldgraf:
cool - just wanted to make sure I wasn't confusing folks :-)

pradyunsg:
We could move the sidebar collapse button to be attached to the side bar's content-side edge, instead of in the topbar?

choldgraf:
@pradyunsg ya something like that might make sense, although then I feel like we are back to square one on the main reason I wanted to play around with this, which is that the top-left of the page is kind of "dead space" right now 🤔

pradyunsg:
Status quo feels fine TBH. I'll probably spend some more time on sphinx-basic-ng this weekend, since that might make it easier to try things for us here.

choldgraf:
Nice! Lemme know if you'd like a pair of 👀 on something in the NG theme

choldgraf:
closing this one since it was mostly an experiment and it doesn't seem like there is a strong demand for a full-width topbar!

Issue #165, opened 2020-08-07
Tie the theme to good design principles


chrisjsewell:
With my scholarly hat on, it would be nice to apply a bit of theory to the design.
It feels mainly the theme is being designing based on "I like the look of that", which is not necessarily a bad thing, but it would be good to do some due diligence and documentation on how the theme applies to good design principles.
Just googling "web page design principles" you get 1000s of results, and obviously there's no silver bullet, but some that start to crop up often are:

- Simplicity and Visual Hierarchy: too many elements on the page can be distracting, the eye should be drawn to the most important
- Consistency: in typographies and color pallete
- F Pattern: Utilising eye tracking software, studies have identified that people tend to view computer screens in a ‘F’ shaped pattern. Meaning that users tend to focus on the left side of the screen and will only take in a few elements on the right side.
- mobile first: its generally easier to design for the simpler mobile first, then add in features for desktop
- Responsiveness/Load times

choldgraf:
I love this idea. Agreed - your assessment of "I like the look of that" is partially true though I've definitely drawn inspiration from some specific other themes and sites (most notably, git book and the tufte css guide)

choldgraf:
A few others that I have used:

- Draw attention to the content. Find ways to "highlight" the center column of the page.
- Mimic other modern book-like structures. Don't do anything too wacky that users aren't used to. Instead take inspiration from some of the modern book themes out there. Some inspiration:
  - [metaflow docs](https://docs.metaflow.org/)
  - [bookdown](https://bookdown.org/yihui/bookdown/)
- Draw inspiration from publishing-specific design and elements
  - in particular, [the Tufte CSS page](https://edwardtufte.github.io/tufte-css/)

pradyunsg:
> Responsiveness/Load times

Throttling myself to "Good 2G" speeds in Firefox's development tools, disabling cache and loading https://sphinx-book-theme.readthedocs.io/en/latest/ today takes... 59s to load the page. :o

The main culprits for this are:

- https://unpkg.com/thebelab@0.5.1/lib/index.js (~500 kB)
- https://unpkg.com/@jupyter-widgets/html-manager@0.18.4/dist/embed-amd.js (~900 kB)

Even without throttling, I can see that my browser waits for a bit downloading from unpkg.com before showing me the page.

pradyunsg:
Here's the screenshot of the dev-tools (full screen, zoomed out to show all assets)

[...]

PR #344, opened 2021-06-11
✨ IMPROVE: Spacing and width defaults for better accessibility

choldgraf:
This PR gives the main content a maximum width so that there aren't too many characters per line. [Web Accessibility Standards](https://www.w3.org/WAI/tutorials/page-structure/styling/#:~:text=Line%20length,characters%20or%20less%20are%20shown.) suggest using a width so that only ~80 characters are displayed per line. This width is slightly larger than that but hopefully not in a problematic way.

If you want to play around with it, try the RTD demo and resize the width to wide/narrow. You should see two margins pop up at some wax width.

related to #165 

# A few references

- [Best practices for line width and accessibility](https://www.w3.org/WAI/tutorials/page-structure/styling/#:~:text=Line%20length,characters%20or%20less%20are%20shown) and [this criterion page from W3C as well](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-visual-presentation.html)
- [Visual design guide from accessibility.digital.gov](https://accessibility.digital.gov/visual-design/typography/)
- [wikipedia page discussing their intent to cap the line width](https://www.mediawiki.org/wiki/Reading/Web/Desktop_Improvements/Features/Limiting_content_width#Goals_and_Motivation)
- [the Tufte CSS page](https://edwardtufte.github.io/tufte-css/)
  - a max-width of 1400px for all content, and then sets 55% for the main text and 45% for the sidebar text.
  - Main content: 55% of full width. Roughly 85 characters per line. Font size 1.4rem
  - Margin content: 50% of width of the main content. Roughly 60 characters per line. Font size 1.1rem
- [webtypography](http://webtypography.net/2.1.2) recommends about 66 characters per line

Note that the max width in this PR is a bit wider than most of the recommendations above. This is partially arbitrary, but partially because of a similar argument to the Wikipedia page above, I could imagine that Jupyter Books are often more information-dense with scannable material, and this might help people scan more easily.

# Demos

New behavior:

Old behavior:


chrisjsewell:
Eurgh, I'm very much against this; you buy a nice wide screen, and then a site decides that you are not allowed to use it 🤷 

Literally those same web-standards you link have full page text!

Also wikipedia doesn't limit the width, which is probably the biggest purveyor of web articles I can think of.

If you really want it, it should be configurable, and off by default

pradyunsg:
> Also wikipedia doesn't limit the width, which is probably the biggest purveyor of web articles I can think of

Wikimedia is doing a redesign right now of their interface, that fixing this is one of their goals: https://diff.wikimedia.org/2020/09/23/wikipedia-is-getting-a-new-look-for-the-first-time-in-10-years-heres-why/

> Our second change introduces a maximum line width to our content on pages where reading is the focus, such as article pages and discussion pages. Research has shown that limiting the width can lead to better retention of the content itself, as well as a decrease in eye strain.

chrisjsewell:
Good for them, but I still want ours to be configurable 😉

choldgraf:
Thanks @pradyunsg for those extra references, I'll add to the top PR to aid discussion as well.

I found [the wikipedia discussion you noted](https://www.mediawiki.org/wiki/Reading/Web/Desktop_Improvements/Features/Limiting_content_width#Goals_and_Motivation) to be a reasonable argument.

I'm +1 on having the max-width be configurable, though I think we should discuss and decide on the default. Our default themes should make good out-of-the-box choices for people based on best practices, so I want to make sure that we have a good rationale for whatever direction we take. This PR was mostly about having something to look at and get feedback. I'm curious what others think.

damianavila:
> I'm +1 on having the max-width be configurable, though I think we should discuss and decide on the default. Our default themes should make good out-of-the-box choices for people based on best practices

I agree with that point of view.
A sensible default is super important and configurability is always welcome 🤗  .

pradyunsg:
First off, I’ll say: I’m completely fine with whether content-width is configurable in SBT or not. I don’t really feel that this theme needs to account for my preferences/biases. :)

TLDR: More configuration options != more good. IMO the only users who would want to change content width, would be the ones who wanna change it to something bad.

I’ve grown to strongly dislike providing configurable content-width to documentation authors in documentation themes. If the theme has picked a reasonable choice, most users won’t want to change it, either because they don’t care or because they understand the importance of content width being limited sanely by default was a factor in picking the theme.

To me, the authors most likely to use this knob would be doing things that just a plain degradation of reading experience, for folks who aren’t like the author or on a completely different setup/workflow. If this knob is being used, it’s not being used to slightly tweak the width; it’s to drastically tweak the width. An example of this, is what virtualenv.pypa.io is doing — which is a realllly bad reading experience in favour of “use all the pixels”.

Conversations with the author, around these design choices are exceedingly difficult to navigate usually since most authors who are doing this need to be convinced that what they’re doing is hurting their documentation readers, and chances are that isn’t particularly a priority if they’re making such a broad change without looking up the TON of literature (and literally every website design that was done with a team of designers) that very clearly communicate “don’t do that”. And that’s assuming the author hasn’t done it because “I like it”. Because now you need to approach it from a different angle. :)

All this is to say, this knob when used, will likely be used to do more harm than good to the reader’s experience. The damage is, of course, soft-of limited to just that documentation set. I personally don’t think that’s acceptable even then, which is why I don’t like providing these knobs.

pradyunsg:
FWIW, An open question would be how should margin content work with this change?

damianavila:
I really like these passionate replies :wink:
You have some really good points about why more configurabilty is sometimes detrimental.
I am pretty sure I could find some examples or biases on the other side (not providing enough config options being detrimental).
Most likely, the key is to find the subtle point where you provide enough configurability to let people build on top of your stuff without the risk of damaging themselves. When that subtle point is not easy to achieve I am probably biased towards providing config options even when it is risky... but in the end, it is my bias 😉 .

choldgraf:
@pradyunsg a couple thoughts below!

re: configurability, one fairly easy way around this might be to document what CSS rule to apply to change this value. That's a little bit harder than a config option, and also doesn't commit ourselves to more testing etc infra.

re: the margin, I think one benefit to provide a max width to the main content is that it could provide more room for margin content. Ours is fairly small right now (I believe about 25% of main content), and if we could get this to 33% without crowding things too much, I think that'd be a more useful use of that space (and would make "full width" content pop out more effectively). This is closer to what the [tufte theme does](https://edwardtufte.github.io/tufte-css/), which is a major source of inspiration for this theme's structure.

choldgraf:
closing this as we can follow up in later discussion and iteration, we've also got some conflicts so it'll probably be easier to start from scratch

PR #346, opened 2021-06-20
🐛 FIX: Path to custom output directory for _sources

dfm:
In #339 I introduced a fix for the `_sources` directory path that works with the default sphinx Makefile. But I have since discovered that this doesn't work with a custom output directory (as used by ReadTheDocs, for example). This updated PR should do the trick more generally. Sorry about the noise!


choldgraf:
oops I missed this PR, sorry! Is there a way that we can confirm this really solves the problem this time? E.g., is there a test that could make it fail w/o this PR, and passes with this PR?

dfm:
No worries at all! I'll do some poking around and see what I can come up with.

dfm:
@choldgraf: ok - I added a test that checks to make sure that the `ipynb` generated for a myst notebook ends up in the right `_sources` directory. It uses subprocess because I couldn't figure out how to pass `outdir` to `sphinx_build_factory`. Happy to update if there is a better way. This fails on `master` but passes with this fix. Thanks!

tests/test_build.py
```diff
@@ -302,3 +302,11 @@ def test_logo_only(sphinx_build_factory):
         "h1", attrs={"id": "site-title"}
     )
     assert not logo_text
+
+
+def test_copy_rendered_notebooks():
+    target = Path("build/outdir/_sources/section1/ntbkmd.ipynb")
+    if target.exists():
+        target.unlink()
+    check_call("sphinx-build -W -b dirhtml tests/sites/base build/outdir", shell=True)
```
choldgraf, line 311:
```suggestion
    # Switch to sphinx_build_factory if we figure out how to use outdir w/ it
    check_call("sphinx-build -W -b dirhtml tests/sites/base build/outdir", shell=True)
```

choldgraf:
Nice - that seems reasonable, I added in a comment to make it clear why we're using `sphinx-build` there. Thanks for this improvement!

Issue #351, opened 2021-07-03
Make our admonitions / code blocks a bit softer


choldgraf:
In some earlier mockups from @DrDrij he showed off some simple ways to make our admonitions and code blocks a bit "softer", and to use a bit less color/"ink" to differentiate elements. Here are two things that stood out to me:

1. Removing the sharp edges + drop shadow from admonitions, and instead using a rounded edge with a faint grey line to differentiate:

   
   <img width="707" alt="Screen Shot 2021-07-03 at 3 39 04 PM" src="https://user-images.githubusercontent.com/1839645/124368370-61502b80-dc15-11eb-8acd-58ec51bc591c.png">

   and

   <img width="690" alt="Screen Shot 2021-07-03 at 3 39 34 PM" src="https://user-images.githubusercontent.com/1839645/124368373-66ad7600-dc15-11eb-96ab-2a3ae881c53b.png">

2. Removing the bounding boxes from the "previous and next" buttons and instead using some bolded text to make them more noticeable.

   <img width="762" alt="Screen Shot 2021-07-03 at 3 40 40 PM" src="https://user-images.githubusercontent.com/1839645/124368375-6ad99380-dc15-11eb-88ce-7e6a15ad8b32.png">

I feel like both of these would be fairly easy to implement, and may be a nice way to soften/flatten some of our content, while also removing the amount of ink and visual elements the user must pay attention to (e.g. the bold vertical line to the left of admonitions).

Curious what folks think about this, I may try whipping together a PR to see what this looks like.

choldgraf:
Another option is that we could **just** set the border radius slightly above 0, so that admonitions and code blocks have a similar rounding to our `sphinx-panels` cards, e.g.:

<img width="823" alt="Screen Shot 2021-07-03 at 5 29 42 PM" src="https://user-images.githubusercontent.com/1839645/124369769-5b157b80-dc24-11eb-94f3-360a919023e9.png">

pradyunsg:
Rounding out the border radius to the same amount as the left border is the approach I took in Furo: https://pradyunsg.me/furo/reference/admonitions/#supported-types

IIRC, I also reduced the shadow on the admonition in Furo; to make things look a little more "tight".

pradyunsg:
Relevant code: https://github.com/pradyunsg/furo/blob/fe0088363a163cd9d6ffaf274560533501e935b5/src/furo/assets/styles/content/_admonitions.sass#L9-L10


```\

[...]

PR #352, opened 2021-07-03
✨ ENHANCE: Soften edges in admonitions and remove some ink

choldgraf:
This is an experiment at implementing #351 

It rounds the corners of our admonitions and uses a grey border rather than a drop shadow. It also makes the "previous/next" buttons use bolded text without a border.

# Tasks

- [x] Get some feedback on these changes
- [x] Convert into "non-draft PR" for final review


cooperrc:
I like the rounded edges, but miss the left hand side bolder color. 
Errors/warnings are great in the squared, bold style, but the muting is much better for notes and other nice-to-know sections.

choldgraf:
The latest push takes @pradyunsg and @cooperrc 's advice and keeps the left border, but rounds the edges and controls the box shadow a bit:

image

choldgraf:
thanks for the feedback all - this is a much more subtle but I think nicer addition :-)

PR #353, opened 2021-07-08
Re-using Furo's reference documentation

choldgraf:


damianavila:
From the description in the first post 😛 , it is not clear to me why you want to add Furo's reference docs here...

chrisjsewell:
Yeh I'm not sure if this is the best approach 😬

pradyunsg:
I believe the useful context is here: https://github.com/pradyunsg/furo/discussions/181

choldgraf:
Yeah sorry for the lack on context. I left it in draft mode to try to avoid signaling that this needed review. I wanted something concrete to point to in the discussion in the furo docs that @pradyunsg linked to

PR #363, opened 2021-08-03
📚 DOCS: Add reference for html_theme_options

### LINKED ISSUE #361: Add summary (table) of html_theme_options
i.e. all the options in https://github.com/executablebooks/sphinx-book-theme/blob/master/sphinx_book_theme/theme.conf (and include their defaults)

If users are anything like me, they can't be bothered to read through all the customisation documentation to find the available keys 😉 

I think this should probably be on the front page, containing links to the particular sections that explain them in more detail

choldgraf:
Adds a reference for all the configuration options that we support. Roughly inspired from [Furo's customization landing page](https://pradyunsg.me/furo/customisation/) but laid out like a traditional table.

closes #361

### DIFF
--- a/docs/customize/index.md
+++ b/docs/customize/index.md
@@ -1,7 +1,36 @@
 # Customization
 
 These sections describe a few ways that you may customize the look and feel of your theme.
-See the sections below for more information.
+
+## Theme options
+
+The following options are available via `html_theme_options`
+
+% Generated from https://www.tablesgenerator.com/markdown_tables#
+% If you make edits, consider pasting there and re-generating the table so it's nicely formatted
+
+| Key                     | Type | Description                                                                                                                                                                 |
+|-------------------------|-------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `single_page`           | bool        | Remove the left sidebar and treat the site as a single page. See [](customize:single-page).                                                                                 |
+| `path_to_docs`          | string      | Path to the documentation, relative to the repository root (e.g. `docs/`). See [](customize:source-files).                                                                  |
+| `repository_url`        | string      | URL of the repository for the documentation (e.g. the GitHub repository URL). See [](source-files:repository).                                                              |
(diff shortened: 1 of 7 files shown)


sphinx_book_theme/theme.conf
```diff
@@ -7,14 +7,16 @@ stylesheet = sphinx-book-theme.27151f2de8a6da6e72b57ec231ceb131.css
 
 [options]
 single_page = False
-expand_toc_sections = []  # DEPRECATE after a few release cycles
+# DEPRECATE after a few release cycles
```
chrisjsewell, line 10: As mentioned at the top of the file, you shouldn’t directly write to this file; it it created in the pre-commit from src/jinja/theme.conf.j2

choldgraf: hmm I thought I had edited the `.j2` file, the diff shows up here: https://github.com/executablebooks/sphinx-book-theme/pull/363/files#diff-f3d11a1ed4be749cbb3f797cd92269c7b9f3fcee90efce84f66c4bfb052dfc5bR10

did I do that wrong?

chrisjsewell: Ah yeh your write, I was accidentally only looking at the pre-commit commit change

docs/customize/index.md
```diff
@@ -1,7 +1,36 @@
...
 
 These sections describe a few ways that you may customize the look and feel of your theme.
-See the sections below for more information.
+
+## Theme options
+
+The following options are available via `html_theme_options`
+
+% Generated from https://www.tablesgenerator.com/markdown_tables#
+% If you make edits, consider pasting there and re-generating the table so it's nicely formatted
```
chrisjsewell, line 10: Maybe consider just using a list-table here

choldgraf: good point

choldgraf:
ahh I just realized why the pre-commit.ci kept changing stuff, I am using a fresh terminal environment and had forgotten to install pre-commit again

docs/customize/index.md
choldgraf: latest push uses list-table

chrisjsewell (approved):

PR #364, opened 2021-08-03
UPDATE: Sphinx v4.0.0

choldgraf:
This PR enables `sphinx>4` to be supported


choldgraf:
any reason not to just merge this @chrisjsewell ?

chrisjsewell:
> any reason not to just merge this @chrisjsewell ?

Yeh should be fine if it’s passing tests, although it’s always good to have a quick look at https://www.sphinx-doc.org/en/master/extdev/deprecated.html and think if anything applies here

choldgraf:
good catch @chrisjsewell - `logo` is deprecated in favor of `logo_url` in V4, so I added a check for this and also noted when to remove it

chrisjsewell (changes requested):
@choldgraf I note that both the tests (https://github.com/executablebooks/sphinx-book-theme/pull/364/checks?check_run_id=3263717973) and document's build (https://readthedocs.org/projects/sphinx-book-theme/builds/14417635/) are still not being actually run with sphinx version 4.

So obviously this isn't yet testing compatibility with v4 😬 and we need to look what is still pinning to v3

choldgraf:
@chrisjsewell ah OK lemme take a look at the test infra, still getting the hang of tox

choldgraf:
Latest commit adds Sphinx 4 to the testing matrix and removes Sphinx 2. Also updates the dependencies to be alphabetical :-)

mmcky:
@choldgraf there are test failures when I upgrade to `myst-nb~=0.13` for testing. I will look more into this tomorrow morning but have reverted the change for now.

setup.py
```diff
@@ -28,13 +28,13 @@
     license="BSD",
     packages=find_packages(),
     install_requires=[
-        "pyyaml",
-        "docutils>=0.15",
-        "sphinx>=2,<4",
-        "click",
-        "pydata-sphinx-theme~=0.6.0",
         "beautifulsoup4>=4.6.1,<5",
+        "click",
```
choldgraf, line 32:
```suggestion
        "click~=7.1",
```

setup.py
```diff
@@ -56,12 +57,12 @@
             "sphinxext-opengraph",
         ],
         "testing": [
-            "myst_nb~=0.11.1",
-            "sphinx_thebe",
             "coverage",
+            "myst_nb~=0.11.1",
```
choldgraf, line 61:
```suggestion
            "myst_nb~=0.13",
```

setup.py
```diff
@@ -28,13 +28,13 @@
...
     packages=find_packages(),
     install_requires=[
-        "pyyaml",
-        "docutils>=0.15",
-        "sphinx>=2,<4",
-        "click",
-        "pydata-sphinx-theme~=0.6.0",
         "beautifulsoup4>=4.6.1,<5",
+        "click~=7.1",
+        "docutils>=0.15",
```
choldgraf, line 33:
```suggestion
        "docutils>=0.15,<0.18",
```

choldgraf:
@mmcky I got all the tests passing :-) could you give a quick review?

docs/conf.py
```diff
@@ -60,7 +60,7 @@
     # "linkify",
     # "substitution",
 ]
-myst_url_schemes = ("http", "https", "mailto")
+myst_url_schemes = ["http", "https", "mailto"]
```
mmcky, line 63: @choldgraf this is now the default in `myst-parser` so we can remove this now.

sphinx_book_theme/_templates/sidebar-logo.html
```diff
@@ -1,7 +1,11 @@
 <div class="navbar-brand-box">
     <a class="navbar-brand text-wrap" href="{{ pathto('index') }}">
       {% if logo %}
-      <img src="{{ pathto('_static/' + logo, 1) }}" class="logo" alt="logo">
+        <!-- `logo` is deprecated in Sphinx 4.0, so remove this when we stop supporting 3 -->
```
mmcky, line 4: @choldgraf good find! Thanks.

mmcky (dismissed):
thanks for finishing this off @choldgraf looking good.

mmcky:
> https://www.sphinx-doc.org/en/master/extdev/deprecated.html

@chrisjsewell that is a great link. Thanks 👍

sphinx_book_theme/_templates/sidebar-logo.html
mmcky: Oh cool it looks like there are `html_logo` and `latex_logo` now

https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=logo#confval-html_logo

This is something we will need to manage in jupyter-book to issue a depcration notice for `logo` in https://github.com/executablebooks/jupyter-book/issues/1438

mmcky:
- [ ] @choldgraf this fails under `sphinx3` against a fixture so I will add in `sphinx3` and `sphinx4` fixtures

sphinx_book_theme/_templates/sidebar-logo.html
choldgraf: we may not need to deprecate it if we are managing it under the hood though!

chrisjsewell:
Guys, FYI it is going to be a bit more of change than this, to support docutils 0.17.

It now uses a bunch of semantic HTML tags: <main>, <section>, <header>, <footer>, <aside>, <figure>, and <figcaption>, instead of e.g. `<div class="section">`

It will break lots of current CSS, like: https://github.com/executablebooks/sphinx-book-theme/blob/43506962d4f90e6fc74d209d72503929900d84e7/src/scss/_page.scss#L66

choldgraf:
@chrisjsewell I think this means that we will need to choose to *only* support docutils 0.17 and up, or else we're going to have to maintain a bunch of CSS that is meant for pre/post 0.17, right?

mmcky:
ah `docutils==0.17`!

Alternatively we could keep `docutils>=0.15,<0.17` and then upgrade `docutils` to support `0.17` separately? 

I don't think there is anything in the stack that uses `0.17` as a minimum yet.

chrisjsewell:
> I think this means that we will need to choose to only support docutils 0.17 and up, or else we're going to have to maintain a bunch of CSS that is meant for pre/post 0.17, right?

I guess firstly I would have a look at what other themes are doing about this, and obviously we likely also require pydata to make updates as well, so at least for now we will likely have to pin to <0.17

You can possibly do e.g. `div.section,section`. to support both; don't know if there is anywhere that could cause issues

mmcky:
https://github.com/executablebooks/sphinx-book-theme/issues/392

sphinx_book_theme/_templates/sidebar-logo.html
mmcky: OK -- if `html` or `latex` isn't specified use `logo` -- I see where you are going

mmcky (approved):

mmcky:
@choldgraf just updated the branch protection rules. I learnt about this yesterday from @chrisjsewell :-)

choldgraf:
As long as we are not creating version conflicts with other parts of the stack then that works for me

mmcky:
> As long as we are not creating version conflicts with other parts of the stack then that works for me

I have been copying `myst-parser` which is `docutils>=0.15,<0.18` so `sphinx-book-theme` would impose `0.16` in preference to `0.17` as I understand it should be able to resolve in the stack. But maybe we should add #392 as a priority item.

choldgraf:
@mmcky anything left of this one or are you just waiting for someone to approve?

mmcky:
@choldgraf just waiting on an `approval` I think this is ready to 🚢

choldgraf:
ok let's ship this one so that we aren't holding everybody back from upgrading to Sphinx 4 :-)

PR #367, opened 2021-08-06
🐛 FIX: Dropdown menus on mobile

chrisjsewell:
This left translation ensures they are visible on small screens.

Before:

image

image

After:

image

image

This left translation ensures they are visible on small screens


choldgraf:
This looks great to me, this has annoyed me forever

choldgraf (approved):

Issue #272, opened 2021-01-03
Reduce the spacing around the display math element.


akhmerov:
**Is your feature request related to a problem? Please describe.**

Check out the spacing below and above a display math element embedded in text:

- Example from the theme docs:

---

![image](https://user-images.githubusercontent.com/2069677/103489879-08473c80-4e18-11eb-95eb-7b3d9ddf8549.png)

---
- Example with more text:

---
![image](https://user-images.githubusercontent.com/2069677/103489887-15642b80-4e18-11eb-9b9b-67bada5df90c.png)

---
I think the spacing between the equation and the text around does not look good—the top margin is way larger than the bottom one.

**Describe the solution you'd like**

Not the best CSS expert here, but I imagine adjusting bottom margin on a `<p>` followed by a mathjax div would do.

**Describe alternatives you've considered**

...some other CSS adjustments.

choldgraf:
+1 from me, I agree it should be symmetric around both. If anybody wants to try a PR I'm happy to review!

akhmerov:
One tricky thing here is that we're dealing with a semantically incorrect DOM here regardless since display math is a part of the paragraph to which it belongs (see also https://github.com/sphinx-doc/sphinx/issues/8166).

choldgraf:
closed via https://github.com/executablebooks/sphinx-book-theme/pull/369

PR #369, opened 2021-08-07
✨ IMPROVE: improve math styling

### LINKED ISSUE #345: Equation label misaligned
**Describe the bug**

Equation labels appear at a different height than the equations themselves

**To Reproduce**

1. Go to https://myst-parser.readthedocs.io/en/latest/using/syntax.html#equation-eqn-best
2. Observe the following alignment:

image

**Expected behavior**

Not a typesetting expert, but I think the label should have the same baseline.

**Environment**

Ubuntu + chrome & firefox (I observe problem on both)

**Additional context**

Not sure what's the best way to approach this, but adjusting `span.eqno` css seems appropriate.

akhmerov:
Closes #345, #272.

I didn't figure out how to set up the dev environment, so the compiled assets aren't updated. I hope CI picks it up.

---

Some motivation:

- #345 follows https://stackoverflow.com/a/42147108/2217463
- #272 is trickier:
  - There is no selector for an element that is followed by another one, only for a preceding one. That's why I apply a negative top margin to display math that follows a paragraph rather than remove the extra margin from a paragraph.
  - AFAIU 1.15rem margin-bottom of a paragraph seems to be inherited from bootstrap, so I can't use any variable for that.


akhmerov:
Doesn't seem to work on either count :(

choldgraf:
What was the sticking point for setting up the dev environment?

akhmerov:
> What was the sticking point for setting up the dev environment?

Nothing I couldn't handle, just the sheer the amount of tools that I don't (yet) regularly use to install and learn, compared to the expected amount of changes (I incorrectly expected this to be an easy fix).

It was also not completely clear what it'd take—I'd need to install pre-commit to compile the assets, that would in turn install the EBP pre-commit hook, and that will bring in require some minification tools (?).

choldgraf:
Are you going off of the instructions on the [contributing page](https://sphinx-book-theme.readthedocs.io/en/latest/contributing.html)?

I think the main thing is just to do:

```
pip install tox    (installs tox)
tox -e docs-update   (builds the documentation from the local copy)
```

that should be all you need to do. If you want pre-commit to run when you commit things, then also do:

```
pre-commit install
```

and it'll run from then on

akhmerov:
Thanks, this worked, and helped me figure out the issues.

akhmerov:
With modifications (here I added an equation without a label as a check):

---

image

---

Previously:

---

image

choldgraf:
this looks great - thanks for improving this!

PR #370, opened 2021-08-07
📚 DOCS: Add sphinx-design and update docs a bit

choldgraf:
This is a small update to take some inspiration from sphinx-design's docs and play around with its directives etc. It updates our documentation to use sphinx-design a bit!


chrisjsewell:
That sphinx design is pretty good isn’t it 😆 one extra think would be to use the logo svgs I created in figma

choldgraf:
Ah good point - will update the logo to SVG when I'm back at the computer next

chrisjsewell:
I updated the sphinx-design landing page a bit: https://sphinx-design.readthedocs.io/en/sbt-theme/, what do you reckon?

choldgraf:
@chrisjsewell ah very cool! I guess that background is a feature flag in the grid directive now or something? Nice :-)

choldgraf:
Latest push adds in SVG logos instead of PNGs (thanks for the Figma link btw, that is super useful). It also removes the title on the landing page, since "book theme" is already displayed to the left and I think this might be a bit less cluttered.

choldgraf:
I think this is good to go - we can iterate from there. thanks @chrisjsewell for the sphinx-design improvements :-)

Issue #329, opened 2021-04-29
Document 'sidebar-search-bs.html' is now `search-field.html` and other PyData theme changes


fmaussion:
After the last update to `0.1.0`, our builds on ReadTheDocs started to fail ([full log](https://readthedocs.org/projects/oggm-edu/builds/13629587/)):

```
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/cmd/build.py", line 280, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/application.py", line 352, in build
    self.builder.build_update()
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 296, in build_update
    self.build(to_build,
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 360, in build
    self.write(docnames, list(updated_docnames), method)
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 534, in write
    self._write_serial(sorted(docnames))
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/builders/__init__.py", line 544, in _write_serial
    self.write_doc(docname, doctree)
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/builders/html/__init__.py", line 611, in write_doc
    self.handle_page(docname, ctx, event_arg=doctree)
  File "/home/docs/checkouts/readthedocs.org/user_builds/oggm-edu/conda/latest/lib/python3.9/site-packages/sphinx/builders/html/__init__.py", line 1061, in handle_page
    raise ThemeError(__("An error happened in rendering the page %s.\nReason: %r") %
sphinx.errors.ThemeError: An error happened in rendering the page activity_sheets/as_gallery.
Reason: TemplateNotFound('sidebar-search-bs.html')

Theme error:
[...]

PR #371, opened 2021-08-08
📚 DOCS: Correct wrong filename for default search field

dmwyatt:
I *think* this is the correct filename.  The text I replaced refers to a non-existing filename.  
If I use the updated filename I get the correct search box otherwise I get a missing template 
error while building.

closes https://github.com/executablebooks/sphinx-book-theme/issues/329


choldgraf:
ah excellent! and this actually closes https://github.com/executablebooks/sphinx-book-theme/issues/329 :-)

thanks!

PR #381, opened 2021-08-18
test translation

chrisjsewell:


docs/conf.py
```diff
@@ -6,6 +6,7 @@
 author = "the Executable Book Project"
 
 master_doc = "index"
+language = "fr"
```
chrisjsewell, line 9:
```suggestion
language = "ar"
```

choldgraf:
I don't see any translations at all in the RTD builds 🤔 am I missing something?

chrisjsewell:
yeh I'm not sure why that is, can you try locally

choldgraf:
I'll try again when I'm at my computer

chrisjsewell:
Its probably because I think you have to set up your project to be in the language in RTD (as well as in the conf.py?): https://docs.readthedocs.io/en/stable/localization.html#single-project-in-another-language

choldgraf:
closing this one as I think this is resolved

Issue #359, opened 2021-07-28
Text in navigation footer (previous, next) in not translated


pfelber:
### Describe the problem

Hi,

The new navigation footer now has "PREVIOUS" and "NEXT" labels. These labels seem to be hardcoded in the css and are not translated when changing the language of the book. While they can be overridden in the CSS, it might be good to have them generated directly in the book's language (I am using "language: fr").

(As a side note, I preferred the old style with "<<", ">>" and the smaller font.)

Thanks,

Pascal

### Steps to reproduce

This problem can be seen in any generated book when setting the language to anything different from English.

### Versions of your packages

```
❯ jupyter-book --version
Jupyter Book      : 0.11.2
External ToC      : 0.2.2
MyST-Parser       : 0.13.7
MyST-NB           : 0.12.3
Sphinx Book Theme : 0.1.1
Jupyter-Cache     : 0.4.2
NbClient          : 0.5.3
```

# Tasks

- [x] Look into the translation function from https://github.com/executablebooks/sphinx-book-theme/pull/214 to re-use that function
- [ ] Confirm that the translation works as expected

chrisjsewell:
Heya, yeh these navigations are inherited from pydata-sphinx-theme:https://github.com/pydata/pydata-sphinx-theme/blob/f0834772d24c962b7d11643114db88615c43674e/pydata_sphinx_theme/layout.html#L4, and so are not currently handled by the translation functionality included in sphinx-book-theme

So either pydata needs to be "fixed" there (cc @jorisvandenbossche) or we need to override that macro

choldgraf:
Ah this was actually a recent change in the CSS to add "previous" and "next" (it was inspired from the Furo docs). This makes me think we might want to change the implementation as I doubt CSS can be translated unless we turn it into a `_t` template, which seems sub-optimal 🤔

chrisjsewell:
Ah ok yeh no that needs to go in the html templates, so it can be translated

chrisjsewell:
@choldgraf did you check that #374 actually works, because I don't think it will; you are using sphinx's internal translation function which does not include PREVIOUS/NEXT, rather than using the translation function I set up in #214

chrisjsewell:
testing in #381

choldgraf:
ahhh in that case you are right - I didn't realize we needed a separate setup for this. I will keep this on my plate in that case and try to get to it when I have time.

choldgraf:
[...]

PR #384, opened 2021-08-22
🐛 FIX: Fix translation for prev/next

choldgraf:
This uses lowercase `previous` and `next` entries, which is the reason that Sphinx wasn't translating them before. Also reverts back to the `_(` translation text since that's the Sphinx standard.

It also sets previous/next to lowercase so they take up less visual attention, per @pfelber's comments

may address https://github.com/executablebooks/sphinx-book-theme/issues/359 if others confirm


choldgraf:
Merging this in so that others can give this a shot

chrisjsewell:
> Also reverts back to the _( translation text since that's the Sphinx standard.

if this works then that is absolutely fine, but I would note that this is not sphinx’s standard for extensions, it is the internal standard for their own documentation and I think not really meant for external use. This is because it is not possible for extensions to extend the available translations (happy to be corrected though)

PR #386, opened 2021-08-24
✨ IMPROVE: Add chevrons to prev/next

choldgraf:
This adds chevrons to previous / next buttons, inspired from @pradyunsg's beautiful Furo theme :-)

image


choldgraf:
Gonna merge this in so that we can cut a release, which will be an improvement on the current behavior (we probably should have held off on that release a little bit TBH but I think this is looking pretty good now)

happy to spot-check improvements in subsequent PRs

PR #397, opened 2021-09-23
✨IMPROVE: Move CSS hash to GET parameter

### LINKED ISSUE #376: Hash our CSS/JS assets with `digest` requests instead of in filenames
# Summary

In https://github.com/executablebooks/sphinx-book-theme/issues/373 we learned that it is difficult to sub-theme this theme because of the hashes in the CSS / JS assets. @pradyunsg shared a solution that has worked for Furo, which uses [Digest headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest) to specify a _specific version_ of the JS/CSS assets. Here's the commit that added it in Furo:

https://github.com/pradyunsg/furo/commit/11dc2eebe65a64cfdecbabbfaa7f75c66fce5a21

We could follow a similar pattern in this theme, which would make it easier for sub-themes to depend on it since they won't need to know the hash ahead of time, they can just specify a CSS file.

# Implementation

We currently use a pre-commit hook to generate hashed CSS, JS, and a `theme.conf` file that addresses these hashed assets. [Here's the asset compilation configuration for our pre-commit hook](https://github.com/executablebooks/sphinx-book-theme/blob/master/web-compile-config.yml). Instead of specifying the hashes in filenames, we could create / append the hashes at Sphinx build time, following a similar pattern to what Furo does in the commit above.

jacobtomlinson:
Closes #376.

This PR makes use of the `hash` filter introduced in https://github.com/executablebooks/web-compile/pull/5 (web compile will fail until this is merged and released) to add the hash to the URL at the end as a GET parameter rather than baking it into the filename.

This means that when the file changes and the hash is updated the cache will be busted, but the file still has a consistent name for referencing in downstream themes.


web-compile-config.yml
```diff
@@ -1,7 +1,7 @@
 web-compile:
   sass:
     files:
-      src/scss/sphinx-book-theme.scss: sphinx_book_theme/static/sphinx-book-theme.[hash].css
+      src/scss/sphinx-book-theme.scss: sphinx_book_theme/static/sphinx-book-theme.css
```
choldgraf, line 4: I'm guessing you looked at this, but is there a reason that this wouldn't work?

```suggestion
      src/scss/sphinx-book-theme.scss: sphinx_book_theme/static/sphinx-book-theme.css?digest=[hash]
```

choldgraf:
Looks like https://github.com/executablebooks/web-compile/pull/5 was merged - once there's a release I can merge this one in, and cut a release of the book theme

chrisjsewell:
Already released as v0.2.2 😉

web-compile-config.yml
pradyunsg: This is the name of the file to generate from the sass file.

choldgraf: yeah but the previous version of it was:

```
src/scss/sphinx-book-theme.scss: sphinx_book_theme/static/sphinx-book-theme.[hash].css
```

so if we just move the `[hash]` to the end wouldn't it get populated as expected?

```
src/scss/sphinx-book-theme.scss: sphinx_book_theme/static/sphinx-book-theme.css?digest=[hash]
```

Maybe I am not understanding something. :-/

chrisjsewell (changes requested):
Cheers, we also need to remove `src/jinja/theme.conf.j2` (and its generation in ` web-compile-config.yml`) and then also change the name in `sphinx_book_theme/theme.conf` (and remove the comment at the top of that file)

web-compile-config.yml
choldgraf: ahh I see we don't actually want the hash in the filename, disregard this comment

chrisjsewell: Oh yeh wait this is not correct, it would be `src/scss/sphinx-book-theme.scss: sphinx_book_theme/static/sphinx-book-theme.css`, but then in the actual template is where the digest needs to be added. Is this possible in the sphinx machinery?

chrisjsewell:
> Cheers, we also need to remove src/jinja/theme.conf.j2 (and its generation in web-compile-config.yml) and then also change the name in sphinx_book_theme/theme.conf (and remove the comment at the top of that file)

Oh wait, sorry I misread the code, yeh I see what's happening now

choldgraf:
it looks to be working to me! here's the relevant HTML from the RTD build:

```
<html>
<body>
<!--StartFragment-->

<link rel="stylesheet" type="text/css" href="_static/pygments.css" />
--
  | <link rel="stylesheet" type="text/css" href="_static/sphinx-book-theme.css?digest=c3fdc42140077d1ad13ad2f1588a4309" />
  | <link rel="stylesheet" type="text/css" href="_static/togglebutton.css" />
  | <link rel="stylesheet" type="text/css" href="_static/mystnb.css" />

<!--EndFragment-->
</body>
</html>
```

chrisjsewell (approved):

jacobtomlinson:
Thanks for pushing to this and getting it over the line for me!

PR #405, opened 2021-09-30
DOCS: Add note of margin CSS classes

choldgraf:
In a recent chat with @jacobtomlinson I realized that we don't explicitly show how classes can be used to put content in the margins. This adds that content in there!


jacobtomlinson (approved):
Woo this is great!

PR #409, opened 2021-10-07
Add support for navigation depth limit

choldgraf:
This PR adds functionality for the `navigation_depth` parameter from the PyData Sphinx Theme - this will remove all toctree items underneath a certain depth, which can significantly speed up the build times.

This is largely sparked by some challenges that the Dask devs ran into with this theme, but I think it will be common with any community that uses this theme and has a lot of API docs. Here's the Dask issue in question:

https://github.com/dask/dask/issues/8227

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -123,6 +123,7 @@ html_theme_options = {
     # "extra_footer": "<a href='https://google.com'>Test</a>",  # DEPRECATED KEY
     # "extra_navbar": "<a href='https://google.com'>Test</a>",
     # "show_navbar_depth": 2,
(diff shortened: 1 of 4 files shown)


choldgraf:
I'm gonna close this one because we resolved it via a combination of documentation + a sphinx-specific plugin rather than exposing a new option in this theme!

choldgraf:
@maksim-0 I closed this one because the original reason it was being open has been solved, but if you think that it's still valuable to be able to control the depth of the sidebar independently, we can re-open and try to get it in

avivrosenberg:
@choldgraf thanks for posting this solution, had the same problem and it solved it 👍🏻

PR #415, opened 2021-10-16
Refactor template to use flexbox

choldgraf:
This is a slight re-factor of our HTML layout. The end-result should be nearly the same, but hopefully with a slightly more sensible layout that is based on flexbox and uses less hard-coded positions and such. The goal is to make the theme behave a bit more predictably, especially for folks that are sub-theming it.

Not quite ready yet, I think there are some more tweaks to make.

# Tasks

- [x] Fix text shift when scroll to bottom (is actually when scrolling to the header w/ longest length)
- [ ] Figure out why margin content sometimes has a scrollbar
- [ ] Figure out why the content sometimes breaks out of the container to the right in the topbar
- [ ] Make the right edge of the topbar and the page content exactly overlapping
- [ ] ...find other problems and fill in here

closes https://github.com/executablebooks/sphinx-book-theme/issues/416


pradyunsg:
The contents sidebar's a bit messed up. 😅 

1. Scrolling to the bottom of the page, results in a weird movement of the "contents" in the top bar.
2. It doesn't have content on some pages. (https://sphinx-book-theme--415.org.readthedocs.build/en/415/)
3. It clips into prose on other pages. (https://sphinx-book-theme--415.org.readthedocs.build/en/415/tutorials/get-started.html)

The sidebar collapse button doesn't change to an arrow when expanded (I think I like this PR's behaviour more).

choldgraf:
welp, I am a bit stuck on this one - for some reason the main container block is *wider than the screen* in some circumstances. If I collapse the sidebar on a narrow-ish screen, then the button to open the sidebar back up disappears off the screen completely:

gh9RXCXpWh

I'm not sure how to solve this one and I've used up my hacking time this weekend on this :-( if anybody else has a pointer I would appreciate any advice.

choldgraf:
I'm gonna close this as I don't have the development time to push it through, and got stuck on getting the CSS to work :-/

PR #420, opened 2021-10-27
Expand left sidebar to the screen edge

kousu:
I'm working on porting a wiki to your (very elegant!) sphinx theme. I tweaked the colour of the navbar:

https://github.com/neuropoly/neuropoly-docs/blob/sphinx-leftbar-bug-demo/_static/theme.css#L2-L4

```css
#site-navigation {
    background: #f3f5f6;
}
```

It works well on a smaller screen:

small

but on a large screen there is an ugly white background to the left of the navbar:

large

This should reproduce the situation for you:

```bash
python -m venv .venv
. .venv/bin/activate
git clone -b sphinx-leftbar-bug-demo https://github.com/neuropoly/neuropoly-docs
cd neuropoly-docs
pip install .[sphinx]
make html
firefox _build/html/index.html
```

I used these versions:

```
$ pip freeze | grep theme
pydata-sphinx-theme==0.6.3
sphinx-book-theme==0.1.6
$ sphinx-build --version
sphinx-build 4.2.0
```

Conversely, Gitbook's similar theme handles this situation elegantly. You can see how it handles at https://www.neuro.polymtl.ca/ ([archive 1](http://web.archive.org/web/20211027033510/https://www.neuro.polymtl.ca/), [archive 2](https://archive.md/tDZPR)). On a small screen it moves the navbar flush to the left:

gitbook - small

but on a large screen it *expands* the navbar, and **the colour comes with it**:

gitbook - large

The trick is that they've added an extra `<div>` inside their outer navbar, and made it `display: flex; justify-content: flex-end`, which has the effect of right-justifying the content.

This patch achieves the same effect for this theme.

On a small screen:

patched - small

On a large screen:

patched - large

I could use some help refining it. I know it has at least one bug, which is that the `topbar` is `position: fixed` so it doesn't respect the new layout and, in particular, the menu collapse button is hidden, and it's probably not that future proof to do this override:

https://github.com/executablebooks/sphinx-book-theme/blob/618cec08a7eaabd74011c98ebc6a7480d78991c4/src/scss/_page.scss#L11-L18


kousu:
> FWIW, if you're looking for GitBook-style sidebars, you might find https://pradyunsg.me/furo/ to be a better fit.

Oh thanks for the tip! We'll definitely consider it. furo doesn't seem to have the Edit On Github button built in like `sphinx-book-theme` though which is one of those unfortunately high-priority features we need. I suppose I could add it in with [`templates_path`](https://pradyunsg.me/furo/customisation/injecting/), which would probably be about on par with the edits I had to do here afterall.

kousu:
So, any thoughts about this PR?

choldgraf:
@kousu - sorry for the slow response here! end of year + holidays totally sucked my time away, and I missed this one.

Agreed that the behavior you're describing is the right way to go. However, I agree that getting the sidebar collapse button to work properly here might be tricky.

@AakashGfude is planning on a refactor of this theme's underlying structure to use a template here: https://github.com/pradyunsg/sphinx-basic-ng . I think that doing this will *also* resolve this issue (or at least, make it much easier to resolve).

So how about we see how that refactor goes, and if it goes well then we can figure out how to fix it w/ the new theme structure. That work?

kousu:
> @kousu - sorry for the slow response here! end of year + holidays totally sucked my time away, and I missed this one.
> [..]
> Could you open up an issue so that this issue is tracked outside of a PR, and we can try to close it with the new theme refactor?

Of course! My too. Happy New Year and all that.

I've opened this up: https://github.com/executablebooks/sphinx-book-theme/issues/455.  I have a workaround in place that suits me for now so I am happy to be very patient that this get done well.

choldgraf:
awesome, thanks a lot :-)

PR #425, opened 2021-11-03
Collapse Github button to a single click, if only one in use.

kousu:
When I theme my site to only use one of the three Github buttons, it takes two clicks (well, a hover + click) to activate:

Screenshot from 2021-11-03 02-42-14

This could be more fluid!

This patch makes the outer trigger button and the actual goal button collapse into one, if there's only one button active.

With two buttons in use:

```python
html_theme_options = {
    "repository_url": "https://github.com/neuropoly/neuropoly-docs",
    "use_fullscreen_button": False,
    "use_edit_page_button": True,
    "use_issues_button": True,
    "use_download_button": False,
}
```

Screenshot from 2021-11-03 02-37-03

but with one:

```python
html_theme_options = {
    "repository_url": "https://github.com/neuropoly/neuropoly-docs",
    "use_fullscreen_button": False,
    "use_edit_page_button": True,
    "use_issues_button": False,
    "use_download_button": False,
}
```

it's much more direct:

Screenshot from 2021-11-03 02-40-39


choldgraf:
I missed all of these PRs @kousu - sorry for missing them! will take a look ASAP, just wanted to let you know it's on my radar now

choldgraf:
OK I took a look at this one, and decided that it'd be better to standardize some of the topbar button behavior in order to do less special casing for all of the buttons in general. There's a PR for this here: https://github.com/executablebooks/sphinx-book-theme/pull/490 

I think that PR will resolve the same issue that this one does, so I think we can close this one if the other is merged

choldgraf:
closing this one as I believe that it is now resolved - @kousu do you want to try this out with the `master` branch and see if it works as expected for you now?

PR #426, opened 2021-11-11
IMPROVE: translation in the footer for Japanese language

KengoTODA:
Hi :wave:, thanks for making this book theme awesome and open!
I'm using this theme with Japanese translation, and want to suggest two changes in the footer:

1. In Japan, usually the word 'copyright' and the copyright statement is not translated. Better to use "Copyright" in the footer.
    * e.g. https://www.ipa.go.jp/, https://www.dic.go.jp/, and https://www.env.go.jp/
2. In Japanese language, "By foo" in this context will be "foo 著", "foo 作", etc. If we want to put the word before the name of author, it is better to use "著者" (author) instead.


choldgraf (approved):
These seem reasonable to me - and I trust your judgment about proper translations more than my own :-)

PR #438, opened 2021-12-09
✨ IMPROVE: Styling print to pdf

AakashGfude:
The following PR aims to change the structure and styling of single page documents when printing using "Print to PDF" functionality of jb. It primarily relies on two css classes for defining structure(showing, hiding elements from the dom) - `onlyprint` and `noprint`.

Output from this PR looks like this:

Screen Shot 2021-12-15 at 2 08 00 pm

[Store code outputs and insert into content.pdf](https://github.com/executablebooks/sphinx-book-theme/files/7716220/Store.code.outputs.and.insert.into.content.pdf)

The above document's content is from https://jupyterbook.org/content/executable/output-insert.html

Will keep updating the output picture and document here, as we push more changes.

EDIT 1: increased the width content, and handled margin directive as shown in the pic above.
cc: @rowanc1 @damianavila @mmcky


mmcky:
thanks @AakashGfude the new `single-page` formatting looks nice and a lot less like a "printed" webpage. 

The link to the `pdf` doesn;t seem to work - just FYI.

mmcky:
Just some thoughts:

1. might it be possible to add an "option" to turn off "margin notes". (in the theme) They often aren't essential to the thread of a page and it might be nice to use the full width of the page. 
2. Can we remove "Print to PDF"?

AakashGfude:
> Just some thoughts:
> 
> 1. might it be possible to add an "option" to turn off "margin notes". (in the theme) They often aren't essential to the thread of a page and it might be nice to use the full width of the page.

Good idea. We can also style the margin to include it in the main content instead of in the side, and stylize it in a way to make it look like a note?

> 2. Can we remove "Print to PDF"?
Working on this.

AakashGfude:
One more drawback is that `background-color` property is not shown unless clicking the checkbox of `background graphics` . 

Ex:
(Left: on, Right: off)
Screen Shot 2021-12-10 at 5 11 38 pm
Screen Shot 2021-12-10 at 5 23 52 pm
Screen Shot 2021-12-10 at 5 12 24 pm
Screen Shot 2021-12-10 at 5 24 38 pm

sphinx_book_theme/layout.html
```diff
@@ -20,7 +20,25 @@
 <main class="col py-md-3 pl-md-4 bd-content overflow-auto" role="main">
     {% block docs_body %}
     {% include "topbar.html" %}
-    <div id="main-content" class="row">
+    <h1 class="onlyprint">{{ pagetitle }}</h1>
```
choldgraf, line 23: Why don't we just wrap all of this in a div like `<div id="jb-print-wrapper" class="onlyprint">`? I feel like that'd be the most obvious to understand and then we only need one `onlyprint` in there

sphinx_book_theme/layout.html
```diff
@@ -20,7 +20,25 @@
...
+    <!-- Table of contents -->
+    <div id="main-content" class="row onlyprint">
+        <div class="col-12 col-md-12 pl-md-5 pr-md-5">
+        <div>
+            {% set page_toc = generate_toc_html() %}
+            {%- if page_toc | length >= 1 %}
+            <div>
+                <h2> {{ translate(theme_toc_title) }} </h2>
+            </div>
+            <nav aria-label="Page">
```
choldgraf, line 33: we probably don't need aria labels and such because this is being printed to PDF, so screen readers won't make sense, right?

sphinx_book_theme/layout.html
choldgraf: If this were the case, could we just put the TOC code:

https://github.com/executablebooks/sphinx-book-theme/blob/master/sphinx_book_theme/topbar.html#L21-L34

in a standalone `html` file, and then import it there and here, so we don't repeat ourselves?

choldgraf (commented):
This is a nice improvement! I have a few thoughts in there about simplifying the codebase a little bit, but in general I think this is close.

The "print to PDF" thing is annoying, though already present in the theme (I didn't notice that before), so I don't think it should block this PR. Though @AakashGfude if you can figure out how to avoid it, that would be fantastic. Note that I believe this is using bootstrap-native tooltips, so maybe those docs will have helpful pointers: https://getbootstrap.com/docs/4.0/components/tooltips/

sphinx_book_theme/layout.html
AakashGfude: @choldgraf there is a bit of difference in classes between the toc in print and non-print.

sphinx_book_theme/layout.html
AakashGfude: @choldgraf these lectures, I reckon, can also be just downloaded as PDF and read on a computer or tablet offline instead of printing it in particular? In which case, screen readers would be important. Not sure, if this HTML tag translates to pdf output as well though.

AakashGfude:
Note that I believe this is using bootstrap-native tooltips, so maybe those docs will have helpful pointers: https://getbootstrap.com/docs/4.0/components/tooltips/

Thank you @choldgraf for the link. Though I could not find any resource to remove it in a boostrapy way. But I did get the element ids. I am detaching them from the DOM just before calling `window.print()` and then reattaching them again for HTML. A bit hacky but works nicely.

sphinx_book_theme/layout.html
```diff
@@ -20,7 +20,27 @@
 <main class="col py-md-3 pl-md-4 bd-content overflow-auto" role="main">
     {% block docs_body %}
     {% include "topbar.html" %}
-    <div id="main-content" class="row">
+    <div id="jb-print-docs-body" class="onlyprint">
```
choldgraf, line 23:
```suggestion
    <!-- Table of contents that is only displayed when printing the page -->
    <div id="jb-print-docs-body" class="onlyprint">
```

sphinx_book_theme/layout.html
```diff
@@ -20,7 +20,27 @@
...
+                <nav aria-label="Page">
+                    {{ page_toc }}
+                </nav>
+                {%- endif %}
+            </div>
+            {{ super() }}
+            </div>
+        </div>
+    </div>
+    <div id="main-content" class="row noprint">
```
choldgraf, line 43: I'm confused - wouldn't this mean that the whole `#main-content` div doesn't show up when printing?

choldgraf (commented):
Just a couple quick comments, this looks nice!

sphinx_book_theme/layout.html
AakashGfude: This #main-content won't, but this one will https://github.com/executablebooks/sphinx-book-theme/pull/438/files#diff-6cf877d43a0155d9e3706cb22f143f326d4283458bfdcac93017b197c4f02e2bR27 . Looking at this now, i should rename the id of `onlyprint` one to something else, as there are two `main-content` ids at one time in this page.

AakashGfude: or maybe convert `main-content` to a `class` name instead of `id`.

choldgraf: hmmm - it seems reasonable to use a unique ID for the print-only header. That seems like a smaller change than converting it from an ID to a class, IMO

sphinx_book_theme/layout.html
choldgraf: I'm not sure either - but if we think it might help folks with a11y issues, let's just leave it in!

sphinx_book_theme/layout.html
AakashGfude: Fair enough. Changed them to individual ids.

AakashGfude:
@choldgraf this looks ready to go I think if there are no further comments.

choldgraf (approved):
I lost track of this one, but thanks @AakashGfude for the improvements! Let's iterate in subsequent PRs if we find other improvements to make.

PR #440, opened 2021-12-26
FIX: Correct Japanese translations

chokkan:
These Japanese translations are weird and incomprehensible.


choldgraf:
many thanks @chokkan - these translations were auto-generated so they would definitely benefit from somebody that understands the languages better.

choldgraf (approved):

PR #448, opened 2022-01-02
MAINT: Simplify toc scrolling behavior

choldgraf:
This simplifies the logic behind our scrolling behavior so that we are relying on a less-opaque and a more performant method for deciding when to hide the Table of Contents.

**Old method**: The way we currently do this is by setting up a callback that fires every time somebody scrolls, calculates the vertical position of every margin element, calculates whether it overlaps with the table of contents, and hides the TOC if they overlap.

**New method**: Uses the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to detect whether an element is on screen. This will only fire when an elements visibility has changed, not on every scroll. This should help performance a decent amount, and is a less hacky way to do things in general.


choldgraf:
Will leave this one open a bit to see if others have thoughts on the slightly new behavior. Maybe @AakashGfude wants to give it a look since he's going to do a deeper dive on the theme infrastructure?

choldgraf:
OK I got the interactionobserver working for the scroll behavior as well, so this PR no longer changes the default style of pages, it only improves the "watching" behavior for margin content. I think that this is good to go

PR #452, opened 2022-01-05
FIX: Anchor links HTML template fix

### LINKED ISSUE #451: Secondary Sidebar Navigation Not Working in v0.1.8
### Describe the problem

Clicking on the 2nd level headers displayed under the secondary sidebar does not take one to the appropriate section of the page.

### Steps to reproduce

1. Go to https://sphinx-book-theme.readthedocs.io/en/stable/index.html#
2. Click on "Reference Pages" in secondary sidebar(or contents bar).
3. In page navigation isn't observed.

### What's going wrong

In #435 we added a second `super()` call, which generates all of the page's content a second time. This means we have duplicated IDs and links for anchors, which is why this isn't working. We need to call `super()` only once per page. Here are the links:

https://github.com/executablebooks/sphinx-book-theme/blob/1f2c69fbae20dfbc18abab6c794e5b0a5f558fe4/sphinx_book_theme/layout.html#L38-L48

choldgraf:
I had a moment to try and fix this, so here's a start. It's not quite finished, but I'm pushing it in case it saves time for @AakashGfude. Feel free to push to this branch to finish it up, or throw this away and start from scratch :-)

This moves the printing TOC into the same div as the main content, so we re-use the table of contents.

The printing seems to work OK, with one exception: there's a horizontal scrollbar on the bottom, though it seems this is present on `master` anyway

image

fixes #451


AakashGfude:
@choldgraf looks good, thanks for starting the PR. There is a bit of misalignment with `Contents` as can be seen from the pic below.
(LEFT: This PR, RIGHT: master)
Screen Shot 2022-01-05 at 1 48 54 pm
Screen Shot 2022-01-05 at 1 49 03 pm

Will fix that, and we should be good to go.

AakashGfude:
@choldgraf have pushed a few CSS styles, i.e, alignment of `Contents`, overlapping in the footer. the horizontal scroll bar is also removed. 
It looks good to me now.

AakashGfude (approved):

choldgraf:
It looks good to me - thanks for pushing the updates! Will merge and cut a quick release.

PR #463, opened 2022-01-14
FIX: Add aria label to sidebar toggle

choldgraf:
As suggested by @pradyunsg (does this look correct to you @pradyunsg ?)


choldgraf:
merging because I think this one is straightforward and I'd like to cut a release, but let me know if this isn't right!

Issue #417, opened 2021-10-19
Switch our build system to use the `sphinx-theme-builder`


choldgraf:
### Description / Summary

Our current setup commits the compiled CSS + hashes directly into our repository. The biggest downside to this is that merge conflicts happen any time somebody merges a PR when another person has a PR open that also modifies CSS or JS.
We're also using [the `web-compile` package](https://github.com/executablebooks/web-compile) to compile our web assets. This is quite lightweight and useful, though we do miss some of the benefit of the NPM ecosystem, which has a lot more functionality that is designed for building web assets.

Recently there has emerged a new theme development tool called [sphinx-theme-builder](https://sphinx-theme-builder.readthedocs.io/). This is a Python package that bundles a version of `nodejs` along with it, so you don't need to go through the hassle of setting up your own node environment. It also has a command line set of tools that let you quickly perform compilation steps via webpack, so you can use the full webpack ecosystem, vendor fonts/packages/etc as you like but without needing to do a bunch of javascript installs. Because the compilation is part of the build step, it means that you do not need to check your compiled CSS/JS assets into the repository, you can do that at build time, so you don't generate merge conflicts whenever you change the JS/CSS compiled assets. Finally, the theme builder is a nice step towards defining a standard sphinx theme structure, which will make it easier to port patterns and share development between the themes that are out there.

### Implementation details

The PyData theme recently switched over to using this new build system, and it has worked out pretty nicely over there. Here's the PR where @pradyunsg did that work:

https://github.com/pydata/pydata-sphinx-theme/pull/514

kousu:
That's great! That's very clever.

pradyunsg:
Alright, at the rate of ~1 hour per day, I've been working on sphinx-theme-builder and Furo's main branch has moved over.

You might be able to copy more than just the asset build system. :)

AakashGfude:
Closing this issue. As this PR https://github.com/executablebooks/sphinx-book-theme/pull/469 does the switch.

choldgraf:
Woohoo! Thanks @AakashGfude for tackling this!

Issue #392, opened 2021-09-02
ENH: Support docutils 0.17.1


mmcky:
### Description / Summary

The `docutils==0.17+` package makes changes to `css`

From @chrisjsewell

It now uses a bunch of semantic HTML tags: `<main>`, `<section>`, `<header>`, `<footer>`, `<aside>`, `<figure>`, and `<figcaption>`, instead of e.g. `<div class="section">` (see https://github.com/sphinx-doc/sphinx/issues/9001)

It will break lots of current CSS, like: https://github.com/executablebooks/sphinx-book-theme/blob/43506962d4f90e6fc74d209d72503929900d84e7/src/scss/_page.scss#L66

### Value / benefit

Enable the use of `docutils==0.17` as the stack get's upgraded

### Implementation details

The question is should we support only `0.17` or previous version of `docutils` as well?

- [x] `docutils>=0.17.1`
- [x] `docutils==0.16`

Aim is to support `docutils>=0.15`

**note** that we probably want to skip `0.17.0` because it has a nasty bug in it, so we should begin with `0.17.1`

### Tasks to complete

- [ ] Allow 0.17 series docutils in our dependencies
- [ ] Ensure that our documentation builds use v0.17
- [ ] Take a careful look at the CSS to make sure that nothing has broken (see https://github.com/sphinx-doc/sphinx/issues/9001 for inspiration of what to look for)
- [ ] For anything that is broken, add additional CSS rules to accommodate both 0.17 AND below
- [ ] Decide if those rules are too complex to be unmaintainable and resolve if needed
- [ ] Ensure that RTD builds use v0.17

pradyunsg:
Is there any specific reason to support `==0.17` of docutils, when there's a newer bugfix (0.17.1) that fixes certain backwards incompatible bugs?

Sphinx 4 had a pin due to those bugs, but that was removed once there was a new release with those bugs fixed (see discussion in https://github.com/sphinx-doc/sphinx/issues/9088).

chrisjsewell:
> Is there any specific reason to support ==0.17 of docutils, when there's a newer bugfix (0.17.1) that fixes certain backwards incompatible bugs?

By `==0.17` he just all versions of 0.17, i.e. going from 0.16 to 0.17 there is a step change in the HTAML output and thus CSS required, so do you write your CSS to handle both output cases (more complexity), or do you drop support for <0.17 (less compatible)
Have you handled this yet in Furo?

[...]

PR #469, opened 2022-01-18
MAINT: Switch to sphinx-theme-builder

AakashGfude:
This PR switches to using [sphinx-theme-builder](https://github.com/pradyunsg/sphinx-theme-builder), following this issue https://github.com/executablebooks/sphinx-book-theme/issues/417

- [x] Use sphinx-theme-builder as the build backend
- [x] Use webpack to compile and build assets. Run stb compile.
- [x] Build docs.
- [x] Run tests.
- [x] Do a thorough cleanout
- [x] Update Docs


noxfile.py
```diff
@@ -0,0 +1,57 @@
+import nox
```
choldgraf, line 1: What was your reasoning for moving to `nox`? I personally prefer `nox` to `tox` as I think it is more useful for actions outside of testing, but wondering why you made the switch as a part of this PR.

package.json
```diff
@@ -0,0 +1,36 @@
...
+    "css-loader": "^3.4.2",
+    "dedent": "^0.7.0",
+    "extract-loader": "^4.0.3",
+    "file-loader": "^5.0.2",
+    "html-webpack-plugin": "^4.3.0",
+    "imports-loader": "^0.8.0",
+    "mini-css-extract-plugin": "^0.9.0",
+    "node-sass": "^6.0.1",
+    "optimize-css-assets-webpack-plugin": "^5.0.3",
+    "pa11y-ci": "^2.4.0",
```
choldgraf, line 19: Some of these we can likely clean out (e.g. we do not test with `pa11y-ci`). Let's shoot for the minimal possible dependencies

choldgraf (commented):
A couple quick thoughts - thanks!

package.json
AakashGfude: yes. need to do a thorough clean out.

noxfile.py
AakashGfude: Similar reasoning. It is more pythony, and we can run python code inside it to aid testing. Since pydata-sphinx-theme was also using it, so jumped at the opportunity 😬 . Making theme package structures more consistent. But happy to continue with tox, if we conclude its better for this project.

choldgraf: Well I am personally a fan of `nox` for the following reasons:

- I find `noxfile.py` is easier to quickly read and understand what's going on, since it looks "like a Python script". This also likely is more accessible to more people that aren't familiar with software development workflows.
- I find it easier to re-use commands (like the install command) since you just write them as python functions

We don't have an official org policy on using tox vs. nox (at least I couldn't find anything in [the development conventions](https://executablebooks.org/en/latest/contributing.html)), so I am fine to try it out here and get feedback from the team over time on whether they like it or not. If worst comes to worst we can always switch back to `tox`.

chrisjsewell: It's not the end of the world, but I am against using nox.
tox is by far the more commonly used package, and declarative configuration is much preferred over python files. Plus I use tox-conda which there is no equivalent plugin system for nox.
tox is the org policy and is implemented in nearly every repo in EBP. nox shouldn't only be used if there is a clear requirement for more flexible configuration.

chrisjsewell: I'd also note that this is undoing #211, where we already had this discussion, and I'd moved the package from nox to tox

choldgraf: Maybe this is overkill, but we could support both for a time? As long as we document each it shouldn't be a huge deal, and maybe an opportunity to compare their use a bit. Those files don't change much 🤷.

This feels like one of those things that there is no "right" answer for 🤔. I don't have any problems with tox so if push comes to shove I'm fine using that, but for some reason nox just feels more user friendly to me, I can't put my finger on it.

chrisjsewell: So are you advocating for `conf.py` over `_config.yml` also 😝 
Again, I don't want to bike-shed, it should be what is most helpful for people actively doing the development on the package. But, I would most certainly block the change, if what on any of the repos I am actually actively working on 😬

AakashGfude: @choldgraf @chrisjsewell Since both seem to do our work just fine. I will just keep tox in this PR, and we can try nox in the later PRs if we want to.

pradyunsg: I'd suggest keeping this change separate. There's a decent amount of baggage in the nox vs tox topic in this project, and it'd be best to keep it separated.

I'm also on board with nox (see pip's development processes) but that's besides the point.

choldgraf: I agree with @pradyunsg - no need to conflate two different changes in one PR that is already touching a lot of stuff.

AakashGfude:
@pradyunsg where do you suggest something like a https://github.com/executablebooks/sphinx-book-theme/tree/master/sphinx_book_theme/translations folder should go in terms of the layout system according to `sphinx-theme-builder` https://sphinx-theme-builder.readthedocs.io/en/latest/reference/filesystem-layout/ .

choldgraf:
I think we could just put it under theme/sphinx_book_theme/translations

AakashGfude:
> I think we could just put it under `theme/sphinx_book_theme/translations`

@choldgraf since technically we are compiling jsons to *.mo files and and `sphinx-theme-builder` technically encourages any assets that need to be compiled to be put in `src/sphinx_book_theme/assets`, so I am a bit confused. Also, I think we should add the creation of *.mo files in webpack instead of committing them?

pradyunsg:
> we are compiling jsons to *.mo files, and sphinx-theme-builder technically encourages any assets that need to be compiled to be put in src/sphinx_book_theme/assets, so I am a bit confused sphinx-theme-builder.readthedocs.io/en/latest/reference/filesystem-layout. Also, I think we should add the creation of *.mo files in webpack instead of committing them?

Hurrah! I wanted to make it easier / explicit for folks to keep compiled assets and uncompiled assets separate -- and I am happy to see that this caught non-CSS/JS stuff here! :)

> where do you suggest something like a master/sphinx_book_theme/translations folder 

I'd suggest leaving them under `theme/sphinx_book_theme/` for now, and then coming back around to cleaning that up in a follow up PR.

---

My heuristic for thinking about such things is: Would it be less work for reviewers if this was in a separate PR? In this case, I think the answer is a yes. 

It's _technically_ where they are on `main` at the moment as well (so... in theory, someone could do `{% include "sphinx_book_theme/translations/.gitignore" %}` in their templates[^1]). We can certainly clean this up in well-scoped follow up PR. It makes reviewing easier, and, gives you a clearer list of "what all did I end up doing". :)

[^1]: Yes, that's why we should move those files out! :)

choldgraf:
I think @pradyunsg has a good intuition here - we should get this PR into a mergeable state ASAP, and we can always use follow-up PRs to clean things up if need be - those won't be nearly as disruptive from a merge conflict standpoint.

AakashGfude:
Yes, totally agree with doing it in follow-up PRs. I guess mostly wanted to bring this issue to light here. Will create a separate issue once this is merged.  

I have moved the `translations` folder to `src/sphinx_book_theme` instead of leaving them under `src/sphinx_book_theme/theme/sphinx_book_theme`. Because to make the translations module importable I had to create empty `__init__.py` in both `theme` folder and in `theme/sphinx_book_theme` folder (make them modules as well). To Make the dotted import possible. 

As we are importing `_convert` python file of translations for testing:
`from sphinx_book_theme.translations._convert import convert_json`

Let me know if I should revert it and this isn't following proper convention.

AakashGfude:
- [ ]  Removed `theme_dev_mode`, config variable, as I think it's not needed anymore, given `sphinx-theme-builder`. https://github.com/executablebooks/sphinx-book-theme/pull/469/files#diff-827e5b7309121354fc6d5411c429a4a07800bc6abee0db2f47b60c400ec4be99L26

AakashGfude:
Have no idea, how to fix the pre-commit.ci failure😢

choldgraf:
Hmm maybe this is related to the "additional dependencies" config? Could we specify the extra dependency in the error message there?

https://github.com/mgedmin/check-manifest#version-control-integration

I wonder if @pradyunsg could weigh in on whether this is somehow related to the sphinx theme builder configuration setup 🎉

pradyunsg:
You don't need check-manifest anymore. That said, that looks like something that's a bit of a Ubuntu mess related to breaking `ensurepip`?

choldgraf:
Is that because we're not checking in programmatically generated assets into the repo anymore? If that's the case then I am +1 on removing it.

chrisjsewell:
> Is that because we're not checking in programmatically generated assets into the repo anymore?

and by extension, what does this mean for `MANIFEST.in`?

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+]
+
+[project]
+name = "sphinx-book-theme"
+description = "A clean book theme for scientific explanations and documentation with Sphinx"
+dynamic = ["version"]
+readme = "README.md"
+
+requires-python = ">=3.7"
+dependencies = [
```
chrisjsewell, line 21: Does this PR now mean docutils 0.17 is supported?

pradyunsg:
That can be removed as well! sphinx-theme-builder determines what files to include by asking git!

https://github.com/pradyunsg/sphinx-theme-builder/blob/10450cfeb65bc3610a6708ee731069c707978fef/src/sphinx_theme_builder/_internal/distributions.py#L33

chrisjsewell:
I guess that's similar to flit 👍

pradyunsg:
Basically, the sdist ends up being the same as a `git archive` or whatever is the equivalent of downloading from the GitHub repository *after* generating the assets.

pradyunsg:
Yup, basically the same as flit. I'm pretty sure I copy-pasted some code for that too.

choldgraf:
(side comment - I think all of these more recent developments in Python packaging are pretty cool, thanks @pradyunsg for all of the work you're doing in that space :-) )

pyproject.toml
choldgraf: I suspect we'll need to change some CSS still, so I'd assume that it won't be supported in this PR and we shouldn't bump that dependency just yet. I'm working on some HTML/CSS refactoring that I think will add support for docutils .17, but will do that as a follow-up to this PR so that we don't change too many things at the same time.

chrisjsewell: Yeh exactly, I was pretty sure that was the case; if the CSS still only covers `div+class`, and not also semantic HTML, then it does not support docutils 0.17

AakashGfude:
> Basically, the sdist ends up being the same as a `git archive` or whatever is the equivalent of downloading from the GitHub repository _after_ generating the assets.

This is pretty cool, I will remove the MANIFEST file.

AakashGfude:
The docs look good, apart from the margin directive: https://sphinx-book-theme--469.org.readthedocs.build/en/469/reference/special-theme-elements.html .
Because the margin element is `aside` instead of `div`. Will add styling for `aside`.

choldgraf:
@AakashGfude the aside thing is broken because this PR increases the docutils version, I think

AakashGfude:
@choldgraf have done a quick fix here, but will open an issue, to check style updates needed for docutils update.

- [ ] open issue to address style changes for docutils update.

choldgraf:
@AakashGfude I think there's already an issue for it, so double check that one first. Though hopefully there's not a ton to change because we've been removing the "div" assumption here and there and just using class names

AakashGfude:
Looks okayish to review. 
I still need to update the docs though.

src/sphinx_book_theme/__init__.py
```diff
@@ -16,7 +10,8 @@
 from sphinx.util import logging
 
 from .launch import add_hub_urls
-from . import static as theme_static
+
+# from . import static as theme_static
```
choldgraf, line 14: should we just delete this?

src/sphinx_book_theme/__init__.py
```diff
@@ -279,21 +231,20 @@ def setup(app: Sphinx):
     # Configuration for Juypter Book
     app.connect("html-page-context", add_hub_urls)
 
-    app.connect("builder-inited", add_static_paths)
-    app.connect("env-updated", update_all)
+    # app.connect("env-updated", update_all)
```
choldgraf, line 234: Same here - can we just remove this?

src/sphinx_book_theme/assets/styles/_footer.scss
```diff
@@ -2,9 +2,9 @@
 * Footer *
 *********************************************/
 footer {
-    padding-left: 0;
```
choldgraf, line 5: Just to confirm - have you changed anything in the CSS files other than either moving them or updating their formatting in a way that doesn't change end behavior? If you *have* changed something, can you briefly describe what you've changed?

tox.ini
```diff
@@ -11,49 +11,40 @@
 # then then deleting compiled files has been found to fix it: `find . -name \*.pyc -delete`
 
 [tox]
-envlist = py37-sphinx3
+envlist = py38-sphinx3
```
choldgraf, line 14:
```suggestion
envlist = py39-sphinx3
```

Perhaps we should stick with the latest supported Python version as the default tox environment?

webpack.config.js
```diff
@@ -0,0 +1,92 @@
...
+        test: /\.scss$/,
+        use: [
+          {
+            loader: "file-loader",
+            options: {
+              name: "styles/sphinx-book-theme.css",
+            },
+          },
+          {
+            loader: "extract-loader",
```
choldgraf, line 70: Could you add a one-line comment for each of these explaining what it does (very briefly)?

choldgraf (commented):
This looks really nice to me! I checked out this PR locally and was able to successfully run `tox -e docs-live` 🎉 

I've left a few (relatively minor) comments throughout, let me know what you think.

I agree that the main thing left to do is document this infrastructure so that others know what's going on. I don't think we want to block this PR on having a ton of docs. How about:

- Provide some basic information about how to build the theme locally with tox
- Provide some basic information about the sphinx-theme-builder
- Provide some basic information about the webpack and JS-specific files that you've added so others know what's going on

You can use the [pydata contributor documentation](https://pydata-sphinx-theme.readthedocs.io/en/latest/contribute/setup.html) for inspiration about what stuff to include.

Once we have those basic pieces there, so that the contributing documentation is "technically correct" even if not totally complete, then I think we should merge this in and start iterating in more focused PRs from there.

src/sphinx_book_theme/assets/styles/_footer.scss
AakashGfude: I have added a few styles, for the margin directive in this commit https://github.com/executablebooks/sphinx-book-theme/pull/469/commits/4fb51549f59c6075ba5e1eb185ded991a244031e , which I pointed out in this comment https://github.com/executablebooks/sphinx-book-theme/pull/469#issuecomment-1019616614. Still need to find the issue which was created for this. I will update that issue once I find it.

tox.ini
AakashGfude: Sounds good

AakashGfude:
> * Provide some basic information about how to build the theme locally with tox
> * Provide some basic information about the sphinx-theme-builder
> * Provide some basic information about the webpack and JS-specific files that you've added so others know what's going on

@choldgraf I updated the contributing guide https://github.com/executablebooks/sphinx-book-theme/pull/469/commits/fb5ed0adfcbd84cb5e658ccc25951334c2e1dfe1 , which briefly touches or just mentions all these points. I reckon we should add more, but we can do it in a follow-up PR?

I will check the translations README, and update it as well.

webpack.config.js
AakashGfude: Done

choldgraf:
Those contributing docs look good to me 👍 - anything else you want to tackle in this PR? Otherwise I'd say we leave it open until tomorrow in case others want to make comments, and merge it in unless there are any objections. This PR touches a lot of stuff so it'd be good to get it in so that we can start iterating.

AakashGfude:
- [ ] Add @choldgraf's email address

AakashGfude:
Sounds good @choldgraf. I think I am done with any edits/additions in this PR.

src/sphinx_book_theme/assets/scripts/index.js
```diff
@@ -1,3 +1,5 @@
+import "../styles/index.scss";
```
choldgraf, line 1:
```suggestion
// Import CSS variables
// ref: https://css-tricks.com/getting-javascript-to-talk-to-css-and-sass/
import "../styles/index.scss";
```

choldgraf:
OK, I took another pass and added in one quick comment, there are likely still some things that we need to clean up. The styles look good to me in general, but there seem to be a few tweaks to make (maybe related to docutils). I will tackle those in the refactor PR (#471) and we can hold off on closing #392 until that PR is merged as well.

I'll merge this PR so that we can start iterating a bit from here.

choldgraf (approved):

pradyunsg (commented):
Hah. I had clicked the wrong button.

docs/contributing.md
```diff
@@ -88,26 +84,38 @@ This repository is a split into a few critical folders:
...
 To do so, follow these steps:
 
-1. **Make your changes in `src/`**. This folder contains all of the SCSS and Javascript that are used in this site. You should edit the files here, and they will be built and included with the site in the `sphinx_book_theme/` folder at build time.
-2. **Install `tox`**.
-
-   ```console
-   $ pip install tox
-   ```
-3. **Build the documentation**. You can use the following `tox` command:
+1. **Make your changes in `src/sphinx_book_theme/assets/`**. This folder contains all of the SCSS and Javascript that are used in this site. You should edit the files here, and they will be built and included with the site in the `sphinx_book_theme/` folder at build time.
```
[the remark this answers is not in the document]
choldgraf: good point - I can take a pass at the contributing docs in general as well (or spot check other changes as needed) in https://github.com/executablebooks/sphinx-book-theme/pull/472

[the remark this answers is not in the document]
AakashGfude: Thanks, @pradyunsg for pointing this out.

Issue #477, opened 2022-02-06
Fullscreen mode doesn't work in Safari


mathbunnyru:
### Describe the bug

**context**
I open https://sphinx-book-theme.readthedocs.io/en/latest/ using Safari on MacOS and press fullscreen button.

**expectation**
I expected Safari go to fullscreen mode.

**bug**
But instead nothing happens.

```browser console
[Error] TypeError: document.documentElement.requestFullscreen is not a function. (In 'document.documentElement.requestFullscreen()', 'document.documentElement.requestFullscreen' is undefined)
	(анонимная функция) (sphinx-book-theme.js:1:1707)
	onclick (latest:382)```
**problem**

This is a problem for people using Safari.


### Reproduce the bug

1. https://sphinx-book-theme.readthedocs.io/en/latest/
2. Open FullScreen

### List your environment

macOS 12.2 under M1
Safari 15.3

Issue #476, opened 2022-02-05
Fix regressions with a few extensions


choldgraf:
### Describe the bug

Our refactor PR may break a few extensions, but we won't be able to investigate until we make new releases of the extensions and manually test out here. Rather than blocking that PR until this is resolved, I'll use this issue to track the things we need to check up on. See below for a running list:

### things to fix

- [x] Add classes that let us select for components and use this to simplify our regression tests and other tests.
- [x] See if our regressions tests can use a single sphinx version
- [x] Release `sphinx-thebe`, `sphinx-togglebutton`, and `sphinx-copybutton` and make sure everything still behaves as expected
- [x] Thebe loading button behaves incorrectly - it looks like `thebe.js` isn't being loaded properly. (go to `reference/thebe` page and initialize it). Fixed in https://github.com/executablebooks/sphinx-thebe/pull/48
- [x] Remove unnecessary special-case CSS for sphinx-togglebutton (try deleting CSS in general and make sure it still looks OK). **NOTE:** I couldn't test this in the PR because the version is capped, but I tested it locally and it looks good now
- [x] Fix the styling of Thebe codemirror cells - we have underlines that don't look good. Fixed in https://github.com/executablebooks/sphinx-thebe/pull/48
- [x] Missing content here: https://sphinx-book-theme.readthedocs.io/en/latest/reference/extensions.html#sphinx-tabs-tabbed-content
- [x] Incorrect margin caption behavior: https://sphinx-book-theme.readthedocs.io/en/latest/reference/special-theme-elements.html#figures-with-margin-captions

PR #471, opened 2022-01-25
[WIP] REFACTOR: Re-work HTML structure and CSS

choldgraf:
This is a refactor of our HTML and CSS, with the goal of improving a number of the UI/UX issues on the book theme, tightening up our white space and sizing, using a more sensible set of CSS classes that will make the theme behave more predictably, and adding more functionality for the in-page TOC on mobile widths.

It is WIP and will need to be rebased once #469 is merged in. Posting here now to track ongoing progress.

Major improvements

- We now use `sticky-top` to place our sidebar and topbar, rather than `fixed`, which means it will flow more naturally if things like banners are placed at the top of the page.
- Sidebar/topbar/toc font is slightly smaller than the main text font to create more space
- There's more space for the sidebar + toc bar
- Increased padding for the main content, and improved padding behavior on wide screens + with sidebar collapsed
- Programmatic definition of our margin size and placement, so that there's less hacky manually-chosen numbers in there
- The right sidebar is now viewable on mobile, via a button to toggle the toc similarly to the left sidebar
- The topbar buttons are now divided into two div groups (left and right) and use flexbox to position and center them nicely

closes https://github.com/executablebooks/sphinx-book-theme/issues/392
closes https://github.com/executablebooks/sphinx-book-theme/issues/459 (I think?)
checks off the last two boxes in https://github.com/executablebooks/sphinx-book-theme/issues/416

PR #466, opened 2022-01-17
MAINT: Update tests to keep single Sphinx version

choldgraf:
This makes a few general improvements to our testing infrastructure:

- updates our testing infrastructure so that we no longer have to conditionally check Sphinx versions
- cleans up some of the ordering of tests to be easier to parse
- adds documentation about how to handle multiple versions of Sphinx in the future

ref #465

PR #478, opened 2022-02-06
MAINT: Update CSS for external packages and remove redundant code

### LINKED ISSUE #468: Get our lighthouse scores to "good" in relevant categories
### Description

In #467 we've added an audit with Lighthouse that will help catch some major regressions in performance, accessibility, etc. The scores for pass/fail are lower than ideal.  We should make the necessary changes to get our lighthouse score up to some minimal threshold.

Deciding on a threshold is somewhat arbitrary, but according to [this guide](https://web.dev/performance-scoring/#color-coding), 90 is considered "good" in a category. To that extent, I think our goal should be to achieve a score of 90 in the "accessibility", "performance", and "best practices" categories.

### Implementation guide

- Some of the scores might be hit by dependencies of this theme, and we could look into making improvements there as well.
- We should consider adding one of the heavier pages (e.g. the notebooks page?) if we want to push this further

choldgraf:
This PR does a few follow-ups to #471 , it:

- Updates some CSS that regressed in #471 
- Updates some of our regression tests so that they're more reliable and don't depend on multiple Sphinx versions (supercedes https://github.com/executablebooks/sphinx-book-theme/pull/466)
- Updates to the latest pydata sphinx theme
- Re-factors our sidebar navigation logic a bit to use less custom python, and to not require a second parse with beautifulsoup

# Todo

- [x] Everything in #476   
      closes https://github.com/executablebooks/sphinx-book-theme/issues/476
- [x] Figure out why the pydata theme 0.8 breaks the column width for bootstrap, see `content-blocks` for ex. 
- [x] Fix safari bug: 
      closes https://github.com/executablebooks/sphinx-book-theme/issues/477 
- [x] See if our accessibility scores improved with the downstream updates.
      closes #468


choldgraf:
cc @mathbunnyru - I think this fixes the safari fullscreen bug. Wanna try it at: https://sphinx-book-theme--478.org.readthedocs.build/en/478/ ?

mathbunnyru:
@choldgraf it half-works :)
So, it goes to the fullscreen, when I press the button in Safari.
But when I press it again, it doesn't go to normal mode and stays full screen (it works well in Chrome).

choldgraf:
ok @mathbunnyru try again!

mathbunnyru:
Seems to work now, thanks!

choldgraf:
OK merging this one in as the next step towards our CSS refactor!

Issue #475, opened 2022-02-02
Put uncompiled assets in translations  in a proper directory adhering to `sphinx-theme-builder`


AakashGfude:
> I think we could just put it under `theme/sphinx_book_theme/translations`

@choldgraf since we are compiling jsons to *.mo files, and `sphinx-theme-builder` technically encourages any assets that need to be compiled to be put in `src/sphinx_book_theme/assets`, so I am a bit confused https://sphinx-theme-builder.readthedocs.io/en/latest/reference/filesystem-layout/. Also, I think we should add the creation of *.mo files in webpack instead of committing them?

_Originally posted by @AakashGfude in https://github.com/executablebooks/sphinx-book-theme/issues/469#issuecomment-1018224756_

PR #480, opened 2022-02-10
Translations folder compilation workflow pattern

AakashGfude:
Have put translation source files in `src/sphinx_book_theme/assets` and destiation files(po, mo) in output static folder.
`_convert.py` as before has the compilation code.

Had to put the command to run the compilation `python src/sphinx_book_theme/_convert.py`, explicitly in all tox commands. 
Would like to know if there is a better way.

Side Note: deleting the output `locales` folder when in the live mode, detects the change but reruns the` stb compile` command only. And stb compile has access only to the run npm build.  Because of this, the translations compilation won't happen automatically, when deleting locales folder from the static directory.

Would love ideas from folks about any other better way to do this. 

fixes https://github.com/executablebooks/sphinx-book-theme/issues/475


choldgraf:
@AakashGfude I think that I figured out how to add the extra compilation step as part of webpack. I pushed a commit that fixes it to this PR, and also opened up a PR in the theme builder to document what worked: https://github.com/pradyunsg/sphinx-theme-builder/pull/17

We still need to fix the tests but I think that this should work!

AakashGfude:
Yes, I marked it for review, then saw the failing tests, and converted it to draft as it was not ready. But also, wanted to see if I am going in the right direction.

AakashGfude:
Thanks, @choldgraf for putting forward this solution. But I am not exactly sure if I am convinced with it. Is webpack a good place to write python code? Will it be evident when reading the code base, to look at all the compilation steps in webpack. 

But yes, it definitely is more robust, someone using `stb` commands directly here instead of tox, can see the compilation as well.

choldgraf:
@AakashGfude I agree that it's a bit wonky to call python from `npm`, but at the same time, I feel like this should be a thing that happens underneath the `stb compile` step rather than as a separate call that we manually add to `tox`. Maybe in the future it'd be possible for `stb compile` to include some other local commands as well? But in the meantime this seems like a reasonable solution as long as we document it for others to notice.

AakashGfude:
> Maybe in the future it'd be possible for `stb compile` to include some other local commands as well? 

Yes, this would be the most ideal solution. 
Cool, I guess as long as it's been made noticeable, should be good to go. 

@pradyunsg ready for review now :)

src/sphinx_book_theme/assets/translations/README.md
```diff
@@ -1,13 +1,13 @@
...
 
-## Structure of this folder
+## Structure of translation files
 
-- `jsons/` contains a collection of JSON files that were originally created with [the smodin.io language translator](https://smodin.me/translate-one-text-into-multiple-languages)
-- `locales/` contains Sphinx locale files that were auto-converted from the files in `jsons/` by `_convert.py`
-- `_convert.py` is a helper script to auto-generate Sphinx locale files from the JSONs in `jsons/`.
+- `src/sphinx_book_theme/assets/translations/jsons` contains a collection of JSON files that were originally created with [the smodin.io language translator](https://smodin.me/translate-one-text-into-multiple-languages)
+- `src/sphinx_book_theme/theme/sphinx_book_theme/static/locales` contains Sphinx locale files that were auto-converted from the files in `jsons/` by the helper script below.
+- `src/sphinx_book_theme/_compile_translations.py` is a helper script to auto-generate Sphinx locale files from the JSONs in `jsons/`.
```
choldgraf, line 10:
```suggestion
### Translation source files

The source files for our translations are hand-edited, and contain the raw mapping of words onto various languages.
They are checked in to `git` history with this repository.

`src/sphinx_book_theme/assets/translations/jsons` contains a collection of JSON files that define the translation for various phrases in this repository.
Each file is a different phrase, and its contents define language codes and translated phrases for each language we support.
They were originally created with [the smodin.io language translator](https://smodin.me/translate-one-text-into-multiple-languages) (see below for how to update them).

### Compiled translation files

The translation source files are compiled at build time (when we run `stb compile`) automatically.
This is executed by the Python script at `python src/sphinx_book_theme/_compile_translations.py` (more information on that below).

These compiled files are **not checked into `.git` history**, but they **are** bundled with the theme when it is distributed in a package.
Here's a brief explanation of each:

- `src/sphinx_book_theme/theme/sphinx_book_theme/static/locales` contains Sphinx locale files that were auto-converted from the files in `jsons/` by the helper script below.
- `src/sphinx_book_theme/_compile_translations.py` is a helper script to auto-generate Sphinx locale files from the JSONs in `jsons/`.
```

choldgraf (approved):
One suggestion for clarifying our translation compilation step, but other than that, this looks good to me as long as nobody out there thinks it's a really bad idea to call `python` from `npm` :-)

choldgraf:
As a quick aside, I quite like the script+json approach that this theme uses for generating the `.mo` and `.po` files. I wonder if @pradyunsg thinks the sphinx-theme-builder could define its own little translation spec, and add some custom compilation to generate those files from a more human-readable format like the JSON one we have here. There are many barriers to adopting translation practices in themes, so anything we could do to help lower that barrier might be impactful. Just a thought!

src/sphinx_book_theme/assets/translations/README.md
AakashGfude: The more descriptive the better. Thanks @choldgraf for improving the doc side of things

choldgraf (approved):
Looks good to me 👍

AakashGfude:
Great. I will leave this for a day in case @pradyunsg has any comments. And then will merge it. Thanks, @choldgraf for all the improvements.

PR #485, opened 2022-02-13
DOCS: Adding more docs to javascript

choldgraf:
This adds some more documentation to our JavaScript functions and also standardizes using the same comment style.


src/sphinx_book_theme/assets/scripts/index.js
```diff
@@ -55,22 +49,50 @@ var toggleFullScreen = () => {
...
+    active_page !== undefined &&
+    active_page.offsetTop > $(window).height() * 0.5
+  ) {
+    navbar.scrollTop = active_page.offsetTop - $(window).height() * 0.2;
+  }
 };
 
 /**
- * Hide the Table of Contents any time sidebar content is on the screen.
+ * Functionality that is triggered before / after printing the page.
```
AakashGfude, line 72: The function itself is just called on button click, so this comment might not be exactly well placed? The text inside the function dwells more into the before / after part of the print function itself.

choldgraf: gotcha - will clarify this 👍

Issue #307, opened 2021-03-18
Improve CSS for API documentation


choldgraf:
In https://github.com/executablebooks/meta/discussions/286 @rscohn2 noticed that our CSS for API docs looks a bit wonky. @chrisjsewell pointed out that the CSS for this is here: https://github.com/executablebooks/sphinx-book-theme/blob/bdcf1065f0fbf92f786082b78c86af7a280651df/src/scss/_page.scss#L232-L234

so we should fix up the CSS there so that this type of API docs looks good!

rscohn2:
I only have basic knowledge of CSS, but if someone could suggest what properties need to change I would like to try to fix this.

pradyunsg:
Do you have examples of this wonky looking pages?

I'm looking at https://sphinx-themes.org/sample-sites/sphinx-book-theme/demo/api.html, and it seems OK to me on mobile.

rscohn2:
I added a custom.css with the following contents to fix the problem. Not sure if this is appropriate as a general fix.

/* A workaround for https://github.com/readthedocs/sphinx_rtd_theme/issues/647
 * Override display for function signatures so that there is spacing between
 * types and arguments */
dt {
    display: table-cell !important;
}

rscohn2:
Thanks for fixing it.

PR #486, opened 2022-02-13
REMOVE: Theme-specific API styling to re-use PyData

choldgraf:
This drops our theme-specific API documentation (adapted from ReadTheDocs) and re-uses the pydata API style. The issues described in https://github.com/executablebooks/sphinx-book-theme/issues/307 had been resolved upstream, but apparently we were not re-using those styles here, and instead were over-writing them!

In that issue, they provided example code that created problematic API docs:

```
.. c:function:: dds_platform *dds_mpi(MPI_COMM comm, int sz)

.. py:function:: foo(a, b)
```

here's a comparison:

OLD:

NEW:

image

closes https://github.com/executablebooks/sphinx-book-theme/issues/307


matteoaletti:
First of all, thank you for the great work.

For some reasons, I'd like to maintain the previous appearance of the documentation.
Do you think is it possible to somehow configure the theme to re-obtain the previous behavior?
Or is it simpler to pin the version of the theme to the 0.2 ?

choldgraf:
Hmmm - you could certainly pin it, but you'd lose any new updates to the theme. Another option is to just copy all the CSS that was deleted in this PR, and add it as your own custom stylesheet

matteoaletti:
Thank you for your answer.

I was able to solve my issue by adding the file as custom stylesheet.
I just had to do an intermediate step: I converted the "_autodoc.scss" file to a "custom.css" file. For this I have used a tool called scss-compile. I don't know if it was a good idea since I am not an expert of these kind of files, but for now it seems to work fine.
Thank you again.

choldgraf:
yep that seems reasonable to me!

PR #489, opened 2022-02-19
MAINT: Fix translation asset generation

choldgraf:
This fixes a minor bug that was causing our translation static assets to land in the wrong spot.

Going to merge this because our compile step is broken without it!


AakashGfude:
@choldgraf , I had put the folder option in the `convert_json` function for the test to run. As `test_locale_convert` has a custom folder path.

choldgraf:
Yep - saw that after and have already added it back in 🙂

AakashGfude:
Wonderful. Thank you @choldgraf

Issue #470, opened 2022-01-19
Pluggable infrastructure for top-bar widgets


chrisjsewell:
Currently, these are essentially hardcoded into sphinx-book-theme, and thus difficult for users to modify/extend.

With https://furo-topbar.readthedocs.io I've started to develop a pluggable "widget" bar for furo, but it should be possible for this to fit into https://github.com/pradyunsg/sphinx-basic-ng

chrisjsewell:
Related issues #462, #449

choldgraf:
Agree that this would be great. Perhaps a generalization of #462. In my opinion all of the "launch button" stuff should be inside of MyST NB and re-used here (and elsewhere).

What are the major components of a widget? Something like:

- An icon
- Text that is optionally displayed next to the icon (at different viewport widths)
- More descriptive that is displayed with a hover
- One of
	  - A link, we redirect when you click it
	  - A JS hook, it fires when you click it
	  - A list of items inside, displayed when you click it or hover (can be the theme's decision)

pradyunsg:
Here's one way to do all of those, as a list of dictionaries that the user can provide in their theme configuration.

```py
[
    # Page
    {
        "label": "Page",
        "page": "page-one",
    },
    # Icon
    {
        "href": "https://github.com",
        "icon": '<svg><use href="#svg-github"></use></svg>',
    },
    # Icon with Label
    {
        "label": "Discord",
        "href": "https://discord.com",
        "icon": '<svg><use href="#svg-discord"></use></svg>',
    },
    # Icon with JS
    {
        "icon": '<svg><use href="#svg-fullscreen"></use></svg>',
        "onClick": "javascript:toggleFullScreen()",
        "tooltip": "Toggle full screen",
    },
    # Dropdown
    {
        "label": "Dropdown",
        "dropdown": [
            {
                "label": "Page",
                "page": "page-two",
            },
        ],
    },
]
```

I'm not sure how this would fit into sphinx-basic-ng (since this sort of thing would need additional Python-size validation + logic, which I don't want to add there -- keeping it limited to only injecting the right variables into templates and not dealing with any incoming configuration shenanigans). I guess, it could provide a validator for such a structure; leaving it upto themes to figure out how to convert this into HTML that works in their spaces?

chrisjsewell:
Guys have you actually read https://furo-topbar.readthedocs.io/? You seem to be quoting back to me what I've already written in there

pradyunsg:
I have.

[...]

PR #490, opened 2022-02-20
ENH: Standardize header buttons

choldgraf:
This standardizes our header button behavior, to automate more things, and ensure standard structure, style, and behavior. The user-facing API shouldn't change at all, this is all under-the-hood stuff. However, this should make it much less work to create a "pluggable" header infrastructure when the time comes (ref: #470)

Major changes:

- We now have macros that create different kinds of topbar buttons, and we re-use these throughout
- The header button logic now creates a list of dictionaries that define metadata for the header buttons - this is then rendered into the proper HTML at the end
- Updated the style of the header buttons so that it doesn't use alternating background colors
- The timing of the tooltips is a bit more delayed now so they don't pop up immediately
- Rename the "topbar" to the "article header", in line with the basic-ng theme and more specific.
- Had to update (and in some cases simplify) a bunch of test infrastructure to accommodate the more standard buttons

supercedes https://github.com/executablebooks/sphinx-book-theme/pull/425

OLD:

chrome_seWZ5sxM4t

NEW:

chrome_kmEDOUCOK9


choldgraf:
Will merge this one in so that we can cut a pre-release and give others a chance to try it out. I took another look and don't think there are any regressions, but something might pop up (if there is something, I suspect it'd be the launch button URL generation)

PR #495, opened 2022-02-20
FIX: Locale conversion tests

choldgraf:
This fixes our locale conversion code + tests


choldgraf:
Merging this in if tests are happy, as it un-breaks our tests and fixes a (minor) bug

PR #508, opened 2022-02-25
FIX: Fix asset hashing/linking for sub-themes

choldgraf:
This improves our asset hashing code so that it makes fewer assumptions about the presence of a CSS hash, and so that it exposes a function that sub-themes could use to hash their own CSS/JS assets! Also adds some docs about this so others can discover it.

### DIFF
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -16,4 +16,5 @@ architecture
 style
 tests
 translate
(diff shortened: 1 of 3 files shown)


choldgraf:
This un-breaks a bug that causes sub-themes to fail, so I'll merge this one if the tests are happy

PR #510, opened 2022-02-27
STYLE: Increase header height on mobile

choldgraf:
Slight adjustment to make the header a bit taller on mobile

### DIFF
--- a/src/sphinx_book_theme/assets/styles/abstracts/_mixins.scss
+++ b/src/sphinx_book_theme/assets/styles/abstracts/_mixins.scss
@@ -30,3 +30,12 @@
     }
   }
 }
+
+/**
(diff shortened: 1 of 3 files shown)


src/sphinx_book_theme/assets/styles/abstracts/_mixins.scss
```diff
@@ -30,3 +30,12 @@
     }
   }
 }
+
+/**
+ * Header behavior on mobile
+ */
+@mixin header-height-mobile {
+  @media (max-width: $breakpoint-md) {
+    height: $header-article-height + 0.5em;
```
choldgraf, line 39:
```suggestion
    height: $header-article-height + 0.75em;
```

PR #513, opened 2022-02-28
MAINT: Pin our github actions to minor versions

choldgraf:
This pins our github actions to minor versions, rather than patch versions, simulating the `~=` syntax we follow for Python dependencies.


choldgraf:
not sure why some actions don't let you pin just to minor versions, but github actions is unhappy and I don't have time to debug this since it's not easy, so just gonna close this

PR #516, opened 2022-03-02
FIX: Always use dropdown with launch buttons

choldgraf:
This fixes the icon image / `<i>` CSS in the header so that they behave properly when displayed by themselves and not as part of a dropdown. It also makes our launch buttons always use a dropdown so that we don't run into issues with logos that clash with the background.

fixes https://github.com/executablebooks/sphinx-book-theme/issues/514

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -108,7 +108,7 @@ html_theme_options = {
     # "repository_branch": "gh-pages",  # For testing
     "launch_buttons": {
         "binderhub_url": "https://mybinder.org",
-        # "jupyterhub_url": "https://datahub.berkeley.edu",  # For testing
+        "jupyterhub_url": "https://datahub.berkeley.edu",  # For testing
         "colab_url": "https://colab.research.google.com/",
         "deepnote_url": "https://deepnote.com/",
         "notebook_interface": "jupyterlab",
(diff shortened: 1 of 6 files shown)


PhilipVinc:
Thanks!

This indeed fixes the issue, however I think 'design'-wise the result is not terribly great:
 - the colab icon is too bright (yellow on white background) and is hard to read in my opinion 
 - it might not be completely obvious to users that this symbol means colab.

 See this for example, and compare it with the more standard 'open in colab' button.
While the old _rocket launcher_ icon does not have an immediate meaning, at least there was a 'colab' string next to the icon in the drop-down.

Screenshot from 2022-03-04 10-07-11

choldgraf:
@PhilipVinc yeah I had the same feeling as well. The challenge is that the header is *only* icons for everything else, no visible text, so I was worried that having the word "Colab" (or other services) would be jarring. Maybe if we made the background a darker grey so that t stands out?

PhilipVinc:
Or maybe can you use something like this 'extended' logo (maybe with a great background as you say), only when it is 'alone'? 

img

choldgraf:
Actually, I wonder if it would be easier to just always keep these launch buttons a drop-down in general...that way the issues if differing designs etc wouldn't be as much of a problem. Do you think that would be annoying?

PhilipVinc:
I'd be in favour of it. 
The idea of removing the dropdown is nice, but it requires designing the icons.

choldgraf:
I think that might be a simpler fix here as well...

choldgraf:
Latest push makes the launch buttons always use a dropdown list, I think that behavior will be more predictable, even though it means people will need to take one extra action to launch an interactive session. Thanks @PhilipVinc for the feedback here :-)

Issue #488, opened 2022-02-18
Make the announcement dismissable


choldgraf:
### Describe the problem/need and solution

**Context**
Our announcement bar is a way for people to make a message more prominent for users of documentation. It is controlled via a `html_theme_options.announcement` configuration. However, there is no way to control its visibility across sessions.

Many announcements have the ability to **dismiss them**, as well as the ability to persist this across browser sessions, so that it doesn't show up anymore until the announcement is updated by the docs author.

### Guide for implementation

**Design inspiration**

Here's how it looks in Bootstrap:

![chrome_Y9hJgtuR8D](https://user-images.githubusercontent.com/1839645/154691214-5bf1509d-6a58-4aa4-adff-11f32bd584b7.gif)

**Implementation guide**

Rough summary:

- If "dismiss and remember" is enabled, then:
- The announcement is **closed by default**, by `display: none`
- A JS function uses `localStorage` to check whether it has been dismissed.
- If it has not been dismissed already, then some JS removes the `display: none`
- If it *has* been dismissed already, then check whether the announcement has been updated since the last dismissal.
	- If so, then follow the step above.
	- If not, then do nothing

And then, use a `dismiss` button to the right of the announcement (like a little X) that when clicked does:

- Add `display: none` to the announcement div
- Create a timestamp for today (e.g. with `const time = new Date().getTime()`)
- Store this timestamp as `localStorage.setItem('timeDismissed', time)`

PR #518, opened 2022-03-06
ENH: Add support for ReadTheDocs popup

choldgraf:
This adds some JS that "places" the ReadTheDocs popup in the sidebar so that it looks a bit nicer on page load. It creates a "top" and a "bottom" of our sidebar, so that things stay at the bottom regardless of the content at the top. Also adds a bunch of CSS to style the ReadTheDocs popup properly.

chrome_AdGNLQl6JI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,6 +137,3 @@ node_modules/
 # Editors
 .vscode/
 .idea
-
-# Programmatically created stuff
-docs/_static/rtd-data.js
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,6 @@ node_modules/
 # Editors
 .vscode/
 .idea
+
+# Programmatically created stuff
+docs/_static/rtd-data.js
--- a/docs/_static/rtd-data.js
+++ b/docs/_static/rtd-data.js
@@ -1,0 +1,8 @@
+// Dummy data for testing ReadTheDocs footer insertion
+// This mimics RTD data for a project that uses both versions + languages
+var READTHEDOCS_DATA = {
+  project: "frc-docs",
+  version: "latest",
+  language: "en",
+  proxied_api_host: "https://readthedocs.org",
+};
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -171,7 +171,9 @@ for ifile in kitchen_sink_files:
 def setup(app):
     # -- To demonstrate ReadTheDocs switcher -------------------------------------
     # This links a few JS and CSS files that mimic the environment that RTD uses
-    # so that we can test RTD-like behavior.
+    # so that we can test RTD-like behavior. We don't need to run it on RTD and we
+    # don't wanted it loaded in GitHub Actions because it messes up the lighthouse
+    # results.
     if not os.environ.get("READTHEDOCS") and not os.environ.get("GITHUB_ACTIONS"):
         app.add_css_file(
             "https://assets.readthedocs.org/static/css/readthedocs-doc-embed.css"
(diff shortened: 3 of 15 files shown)


pradyunsg:
FYI: I’m not able to see the injected button on an iPad in landscape mode.

chrisjsewell:
Cross-linking to: https://github.com/pradyunsg/furo/discussions/359 (was hoping @pradyunsg would do it, so then we could steal he's implementation 😅)

chrisjsewell:
For me (macbook, firefox), its only visible after scrolling down

choldgraf:
OK the latest commit cleaned things up a bit, but there's still one major UX issue that I think is related to both of the things you reported above:

The height of the sidebar container is `100vh`. However, with the _announcement bar_ at the top, it means that the sidebar extends slightly below the screen. This causes the RTD button to be hidden until somebody scrolls down enough to hide the announcement bar.

A better approach would be to somehow get the sidebar container to stretch to fill the remaining vertical space that fits within the screen, however much it is. However, I can't for the life of my figure out how to do it. I've tried `align-self: stretch`, but that doesn't seem to work, so I'm sort-of at a loss.

chrisjsewell:
@choldgraf, it has actually been implemented in furo now 🎉 , did you look at that implmentation?

pradyunsg:
If it's in a flex container, `flex-grow: 1` will make it fill the parent. I'm not sure if that's what you want tho. :)

choldgraf:
yeah I feel like the issue here is some kind of CSS bug that is probably not directly portable to the Furo theme :-/  I can have another look at how the sidebar behavior works though in case that helps illuminate things.

The weird thing is that on the "mobile view" of my browser, the RTD button shows up in the sidebar just fine, but when I try it with my actual phone, it is hidden below the screen. I dunno what's up with that.

@pradyunsg the issue w/ `flex-grow` is that the flex direction is row, not column, so `flex-grow` makes the sidebar take up more horizontal space, while I want it to take up all of the available _vertical_ space.

pradyunsg:
Ah, the problem is the announcements.

I have the same problem on https://furo.readthedocs.io/en/latest/; so if you figure something out, lemme know. :P

choldgraf:
OK so at least on mobile I think that I've resolved the issue. Apparently `vh` units behave differently across mobile devices and it is less reliable in general, so working around that in the latest commit.

At this point the expected behavior is:

- On widescreen, the RTD button should be visible by default
- On widescreen with an announcement, it will be hidden until you scroll past the announcement
- On mobile, it should be visible when you slide out the left drawer

The second bullet is definitely sub-optimal, but hopefully not a huge deal since announcements will likely be used relatively rarely, and if we can resolve #488 , then they can be dismissed on subsequent pageviews.

choldgraf:
(If anything here seems to be broken _other than_ the announcement pushing down the RTD button, please let me know!)

choldgraf:
Actually I think that I might have solve the issue of the footer floating to the bottom in [executablebooks/sphinx-book-theme@`ddd5d48` (#518)](https://github.com/executablebooks/sphinx-book-theme/pull/518/commits/ddd5d48f213a7d6f985aca4217a0934240eb7270) , we should double-check that nothing else unexpectedly changed.

choldgraf:
OK the latest push cleans up the CSS a bit so that it looks more in-line with how the other visual elements of the theme look. It is maybe a bit harder to spot, but I think that it looks a lot cleaner, and since it's in the same position as the RTD theme switcher I'd hope it won't be too confusing. I updated the top comment w/ a GIF of how it looks.

choldgraf:
ok gonna merge this one in and we can iterate on more improvements if there are issues that pop up!

PR #523, opened 2022-03-10
A simple sidenote, marginnote functionality

AakashGfude:
**EDIT**: Please refer to the other PR for "footnotes vis sidenotes": https://github.com/executablebooks/sphinx-book-theme/pull/546

So, I wrote up a simple sidenote and marginnote role here. Trying to mimic https://edwardtufte.github.io/tufte-css/ . Note that we can't have references inside sidenotes, as it is a role and not a directive.

As discussed in  https://github.com/executablebooks/jupyter-book/issues/598

Screen Shot 2022-03-16 at 11 57 25 am

For smaller screens it will look like this (for sidenote and marginnote):

https://user-images.githubusercontent.com/6542997/158496105-97921c63-bc31-49fd-aba9-15100664dd56.mov

I will keep the PR open for review though, even though it's not ready, for people to deliberate on whether the code is getting bloated here and we should shift it to a new extension. 

cc: @choldgraf @mmcky

### DIFF
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -84,6 +84,17 @@ but I'll stop here.
 ```
 ````
 `````
+## Examples of marginnote and sidenotes
+
+One of the most distinctive features of Tufte’s style is his extensive use of sidenotes{sidenote}`This is a sidenote`.
+Sidenotes are like footnotes, except they don’t force the reader to jump their eye to the bottom of the page,
+but instead display off to the side in the margin. Here is another sidenote{sidenote}`This is a second sidenote`.
+
+If you want a sidenote without footnote-style numberings, then you want a margin note.
+{marginnote}`This is a margin note. Notice there isn’t a number preceding the note.` On large screens,
+a margin note is just a sidenote that omits the reference number.
+This lessens the distracting effect taking away from the flow of the main text,
+but can increase the cognitive load of matching a margin note to its referent text.
 
 ## Sidebars
 
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -84,32 +84,8 @@ but I'll stop here.
 ```
 ````
 `````
-## Examples of marginnote and sidenotes
 
-One of the most distinctive features of Tufte’s style is his extensive use of sidenotes{sidenote}`This is a sidenote`.
-Sidenotes are like footnotes, except they don’t force the reader to jump their eye to the bottom of the page,
-but instead display off to the side in the margin. Here is another sidenote{sidenote}`This is a second sidenote`.
-
-If you want a sidenote without footnote-style numberings, then you want a margin note.
-{marginnote}`This is a margin note. Notice there isn’t a number preceding the note.` On large screens,
-a margin note is just a sidenote that omits the reference number.
-This lessens the distracting effect taking away from the flow of the main text,
-but can increase the cognitive load of matching a margin note to its referent text.
-
-## Sidebars
-
-There are two different kinds of sidebar-like content in `sphinx-book-theme`,
-typical `{sidebar}` directives, as well as a theme-specific `{margin}` directive.
-This section covers both. Both allow you to place extra content
-separately from your main content.
-
-```{tip}
-Sidebar content will generally overlap with the white space where your site's
-table of contents lives. When the reader scrolls sidebar content into view, the
-right TOC should hide itself automatically.
-```
-
-### Margin content
+## Margin content
 
 You can specify content that should exist in the right margin. This will behave
 like a regular sidebar until the screen hits a certain width, at which point this
@@ -117,7 +93,7 @@ content will "pop out" to the right white space.
 
 There are two ways to add content to the margin: via the `{margin}` directive, and via adding CSS classes to your own content.
 
-#### Use a `{margin}` directive to add margin content
+### The `{margin}` directive - block-level margin content
 
 The `{margin}` directive allows you to create margin content with your own title and content block.
 It is a wrapper around the Sphinx `{sidebar}` directive, and largely does its magic via CSS classes (see below).
@@ -134,7 +110,52 @@ Here is my margin content, it is pretty cool!
 ```
 ````
 
-#### Use CSS classes to add margin content
+### The `{marginnote}` and `{sidenote}` roles - inline margin content
+
+In addition to block-level margin content, this theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/){sidenote}`For example, this is a sidenote`.
+
+Sidenotes are numbered, and behave like footnotes, except they live in the margin and don’t force the reader to jump their eye to the bottom of the page.
+On narrow screens, sidenotes are hidden until a user clicks the number.
+If you're on a mobile device, try clicking this sidenote{sidenote}`This is a second sidenote`.
+
+Marginnotes are not numbered, but behave the same way as sidenotes.
+On mobile devices a symbol will be displayed that will display the marginnote when clicked.
+For example{marginnote}`This is a margin note. Notice there isn’t a number preceding the note.`.
+
+
+### Figure captions in the margin
+
+You can configure figures to use the margin for captions.
+Here is a figure with a caption to the right.
+
+```{figure} images/cool.jpg
+---
+width: 60%
+figclass: margin-caption
+alt: My figure text
+name: myfig5
+---
+And here is my figure caption, if you look to the left, you can see that COOL is in big red letters. But you probably already noticed that, really I am just taking up space to see how the margin caption looks like when it is really long :-).
+```
+
+And the text that produced it:
+
+````
+```{figure} images/cool.jpg
+---
+width: 60%
+figclass: margin-caption
+alt: My figure text
+name: myfig5
+---
+And here is my figure caption, if you look to the left, you can see that COOL is in big red letters. But you probably already noticed that, really I am just taking up space to see how the margin caption looks like when it is really long :-)
+```
+````
+
+We can reference the figure with {ref}`this reference <myfig5>`. Or a numbered reference like
+{numref}`myfig5`.
+
+### CSS classes for custom margin content
 
 You may also directly add CSS classes to elements on your page in order to make them behave like margin content.
 To do so, add the `margin` CSS class to any element on the page.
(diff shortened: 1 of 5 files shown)


choldgraf (commented):
Nice! This looks like a good start to me! I think that we could develop the functionality of the roles / directives in the book theme, and then when it's a nice self-contained piece, we could move it into its own extension (except for the CSS, much of which would probably remain here except for some basic functionality).

A few extra thoughts and suggestions:

- are you thinking of having both both a `marginnote` and a `sidenote` role?  I think for roles it makes sense to have both. For directives, I think a `{margin}` directive is probably the only thing we need because it doesn't make sense to give a number to a block-level thing.
- for CSS, we currently are using the Sphinx `sidebar` CSS behavior on mobile devices. Another approach would be to use a `<label>` and `<input>` combo somehow to make the side/marginnotes "toggle-able" like in the Tufte demo. What do you think about that?
- Could you add the use of this extra functionality to our docs somewhere? E.g. a section in our kitchen sink that shows off all the ways the functionality works? That way it's easier to see it in action!

AakashGfude:
>     * are you thinking of having both both a `marginnote` and a `sidenote` role?  I think for roles it makes sense to have both. For directives, I think a `{margin}` directive is probably the only thing we need because it doesn't make sense to give a number to a block-level thing.

@choldgraf  Yes, my thoughts exactly.

>     * for CSS, we currently are using the Sphinx `sidebar` CSS behavior on mobile devices. Another approach would be to use a `<label>` and `<input>` combo somehow to make the side/marginnotes "toggle-able" like in the Tufte demo. What do you think about that?

I have implemented this using node etc. There might be other ways. But it works now. Html is equivalent to the Tufte one. Have updated the top description with a clip of it.

>     * Could you add the use of this extra functionality to our docs somewhere? E.g. a section in our kitchen sink that shows off all the ways the functionality works? That way it's easier to see it in action!

I have pushed a small demo of it in the notebooks.md file https://github.com/executablebooks/sphinx-book-theme/pull/523/files#diff-437767fba3ba2d9d938a05bc82ede5fbeb2f97fdec99c051ff88350b6924ed3cR272 . Will write a proper doc about it, once we figure out if implementing this feature here is a good idea.

AakashGfude:
Have added a marginnote as well, and updated the description.

AakashGfude:
In terms of functionality implementation for HTML based on https://edwardtufte.github.io/tufte-css/, this PR is pretty much complete. Have not taken care of the overlapping of right toc yet now. Latex implementation requires design decisions.

If we plan to just implement this feature here, then I will take care of:

- [ ] tests
- [ ] right toc overlapping
- [ ] pdf print handling
- [ ] latex handling

src/sphinx_book_theme/nodes.py
```diff
@@ -0,0 +1,48 @@
...
+        add_node(
+            cls,
+            override=True,
+            html=(visit_SideNoteNode, depart_SideNoteNode),
+            latex=(visit_SideNoteNode, depart_SideNoteNode),
+        )
+
+
+def visit_SideNoteNode(self, node):
+    if isinstance(self, LaTeXTranslator):
```
choldgraf, line 25: Maybe this should be "if `not` <html translator of any kind>`"? I assume we'd want other non-HTML builds to not include these either right?

docs/content-blocks.md
```diff
@@ -84,6 +84,17 @@ but I'll stop here.
 ```
 ````
 `````
+## Examples of marginnote and sidenotes
```
choldgraf, line 87: could you move this underneath "sidebars" since that's the section that covers things like margin notes?

src/sphinx_book_theme/assets/styles/content/_margin.scss
```diff
@@ -38,6 +38,47 @@ $content-fullwidth-width: percentage(100% / $content-max-width);
...
+      clear: both;
+      width: 95%;
+      margin: 1rem 2.5%;
+      position: relative;
+    }
+  }
+}
+
+span.sidenote,
+span.marginnote {
```
choldgraf, line 74: Note that we should add these to the list of classes the intersectionobserver is watching:

https://github.com/executablebooks/sphinx-book-theme/blob/227e1995652a3d09aec426d13104a6e430696bd2/src/sphinx_book_theme/assets/scripts/index.js#L133-L141

choldgraf (commented):
This looks really nice to me! In my opinion, it is close to good-to-go. I'd be fine getting this one to a place where it can be merged in here. If we want to have it as a standalone package, that could be done in a follow-up. Does that make sense?

docs/content-blocks.md
choldgraf: I'll juts give this a shot

src/sphinx_book_theme/nodes.py
AakashGfude: At present, I am calling this only for HTML, latex builds https://github.com/executablebooks/sphinx-book-theme/pull/523/files#diff-c428944db19a77ff68c5777783e4a45aa9753788f3620049d9128cab09232611R19, so other builds won't be accessing this function I reckon.

AakashGfude:
In the case of latex, should we just leave the text expanded as we have in the case of mobile screens?:

Screen Shot 2022-03-16 at 2 55 13 pm

choldgraf:
Another quick thought, to think about this from a design / UX perspective since I think it's most important to figure that out early on.

There were two other discussions in related issues around this:

- https://github.com/executablebooks/jupyter-book/issues/598#issuecomment-630290544
- https://github.com/executablebooks/jupyter-book/issues/601

One ideas was **standardizing on a single word** (e.g. `{margin}` or `{sidenote}` with some way to configure "numbered or not"). AKA, we don't need verbs for `margin`, `marginnote`, and `sidenote` because that's too much stuff for people to remember. Would it be difficult to implement something like this? That could be a conceptually much simpler implementation.

Another idea was to **re-use the footnote functionality**, and somehow make footnotes sometimes behave like margin content (either via config, a directive, etc.

Curious what you and others think about those patterns from a user perspective - perhaps that could either simplify the implementation here or for the users?

choldgraf:
### Example from pandoc-sidenote

Quick follow-up here, I was looking into how Pandoc does this, and [the pandoc sidenote plugin](https://github.com/jez/pandoc-sidenote) also over-loads the `footnote` syntax. Basically, if that plugin is installed, then it will check the *content* of each footnote on a page, and if it begins with a special character, will do different things.

For example, here's a paragraph that contains a footnote, a side note, and a margin note:

```
My paragraph has a footnote[^fn], a numbered sidenote [^sn], and a non-numbered margin note [^mn].

[^fn]: {.} This is a footnote. The {.} is what causes it to be a footnote.
[^sn]: This is a sidenote it is the default behavior.
[^mn]: {-} This is a margin note without a number. The {-} is what causes it to be a non-numbered margin note.
```

I think the definition of characters it uses is here (in haskell, but you get the idea):

https://github.com/jez/pandoc-sidenote/blob/c9dfcadfcff475e6e26626af379b8b6d3f38d187/src/Text/Pandoc/SideNote.hs#L21-L30

### A proposal

What if we took the same approach, change the default behavior. The default would be "footnotes", and we could check the content of each note for the symbols `{>}` == side note, or `{.>} == non-numbered margin note.

So the behavior from a user's perspective would be:

```
My paragraph has a footnote[^fn], a numbered sidenote [^sn], and a non-numbered margin note [^mn].

[^fn]: This is a footnote. It is the default behaior.
[^sn]: {>} This is a sidenote. The {>} is what causes it to be a sidenote
[^mn]: {.>} This is a margin note without a number. The {.>} is what causes it to be a non-numbered margin note.
```

I think that the implementation here could largely be re-used, but the thing you'd need to do is parse the footnotes in a transform, and replace them with sidenote / marginnote nodes depending on the content of each (and potentially re-position them in the document so they are next to their refs.

What do you think about that from a UX perspective?

chrisjsewell:
> What if we took a similar approach. 

Yep, this is pretty much what I had in mind in our Slack chat 👍 

The idea is that, semantically, they are all basically the same thing: the user wants text that is outside the normal flow of the documentation content (i.e. an [aside](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)).
Different output formats / sphinx HTMLthemes, etc, would then decide the best way to achieve this visually

AakashGfude:
> One ideas was **standardizing on a single word** (e.g. `{margin}` or `{sidenote}` with some way to configure "numbered or not"). AKA, we don't need verbs for `margin`, `marginnote`, and `sidenote` because that's too much stuff for people to remember. Would it be difficult to implement something like this? That could be a conceptually much simpler implementation.

It won't be difficult to implement I think, and conceptually would be sensible as well. But, how should we pass options to roles,i.e, to make it numbered, unnumbered? The only way I see is using syntax like "{sidenote}` text <option>`". Can we have same verb for roles and directives?

> Another idea was to **re-use the footnote functionality**, and somehow make footnotes sometimes behave like margin content (either via config, a directive, etc.
> 

I am not entirely sure on this idea, as the syntax of footnotes are a bit different in that you define a label first. and then define the content of the footnote separately. But we have only one role definition for marginnote and sidenote. Feels like we might need to adjust the functionality to cater marginnote and sidenote enough to maybe write its own code?

AakashGfude:
@choldgraf I have a few questions. 
So, when we use footnote style definitions, we will have content definition and superscripted number reference separately right? like:

```
- This is an auto-numbered footnote reference.[^myref]

[^myref]: This is an auto-numbered footnote definition.
```

If these two definitions are two far away from each other, then we will have to bring them together(which can be messy i think) when the screen size is small, for clickable checkboxes. Ideally, they should be in the same place imo. 

In the case of 
```
[^mn]: {.>} This is a margin note without a number. The {.>} is what causes it to be a non-numbered margin note.
```
why do we need an extra `{.>}` as `[^mn]` itself should imply non-numbered margin?

> * If footers + sidenotes were used at the same time, the numbering would continue across them which would reduce confusion.

I think we should have separate numberings for both of them, as they are two different things. They would be distinguishable given their design. But having numbers continue between them will create confusion between them imo.

choldgraf:
Thanks for thinking this through a bit w/ me - here are some quick responses:

> If these two definitions are two far away from each other, then we will have to bring them together(which can be messy i think) when the screen size is small, for clickable checkboxes. Ideally, they should be in the same place imo

Yep I agree - I think that if we use this footnote notation for them, we'd need to do something like the following:

- In a transform, scan for all the "footnote" nodes in the doctree
- For each one, scan its content for `{>}` or `{.>}`
- If found, convert it into a sidenote or marginnote node
- Insert the "content block" for it right next to its reference node. (I assume the doctree will have a node both for the reference and for the content, since that's how it links them together)

> why do we need an extra {.>} as [^mn] itself should imply non-numbered margin?

In this case the `mn` is just an arbitrary label, right? I believe footnotes behave like:

```
Some text and a footnote with an arbitrary label[^my-arbitrary-label].

[^my-arbitrary-label]: The footnote text.
```

So in this case, `mn` is just the arbitrary label name. That's why you'd need to look at the text block to know if it was a footnote/marginnote/etc.

> They would be distinguishable given their design. But having numbers continue between them will create confusion between them imo.

I think on this one, I agree w/ @chrisjsewell that, while they are structurally in different places (one in the bottom, one to the side), they are semantically the same (they are both "aside" content that is not meant to disrupt the flow of the main content). I worry that having two numberings will be confusing. For example, consider content like:

```
Here's a {sidenote}`Sidenote`, and here's a footnote[^1], which is which?

[^1]: The footnote text.
```

This would render as the following, which seems confusing to me:

image

Would also love thoughts from @rowanc1 @fwkoch or @stevejpurves in case they've thought through markdown/sidenote syntax in curvenote at all

choldgraf:
Actually - as a first step, perhaps the simplest approach would be to avoid creating any new syntax, and for now, just give users the option of turning *all* of their footnotes into sidenotes. AKA, have a configuration option like:

```python
html_theme_options = {
  "use_sidenotes": True,
}
```

at a global level, as well as supporting this config on a page-metadata level for over-rides.

And this would cause all footnotes to behave like sidenotes.

If we wanted the ability to have non-numbered sidenotes, this could be another flag. like "number_footnotes" at a global and page level, and this would remove the numbers for all sidenotes.

This kind-of assumes the following, but I think it's a reasonable assumption:

1. Within a page, people will want either sidenotes, or footnotes, but not both.
2. Within a page, people will want either numbered side/footnotes, or non-numbered, but not both.

Then in the future, we could always add more complex syntax / roles / etc if people asked for it, but for now we wouldn't need to expand the markdown needed at all, and just re-use footnote syntax.

rowanc1:
I think another example is distill.pub (e.g. [here](https://distill.pub/2021/understanding-gnns/#d-footnote-1)) which uses both "hover-notes" (similar to margin notes in that they are close to where you want them) *as well as* numbered footnotes that are at the bottom.

So I think that the following is not unreasonable:
```py
html_theme_options = {
  "use_sidenotes": True,
  "use_footnotes": True,
}
```

I do like the page/book level config though! And simplifies syntax decisions that we can bring in down the road if we need that level of specificity!

AakashGfude:
This sounds good. Just a couple of quick questions:

1) Will this have any effect on the margin directive?
2) Are footnotes used as a place for citations as well? In that case, will the author want to use both footnotes for citations and sidenotes as places to put relevant add info about the text? (Or they can use footnotes for citations and margin directive for add info)

chrisjsewell:
> Are footnotes used as a place for citations as well?

No these are definitely separate

chrisjsewell:
> Will this have any effect on the margin directive?

Potentially margin could be changed to simply generate a footnote, then eventually deprecated?

choldgraf:
Agreed that citations are a totally different thing from footnotes 👍 

I think a starting point would be getting the "sidenotes via footnotes syntax" functionality working, and seeing how awkward it feels to use this for block-level elements.

e.g., how awkward from a UX perspective would it be to turn this:

````markdown
Here's a paragraph with a footnote[^1].

::::{margin}
And here's a longer margin block-level element...
:::{note}
...with a note inside.
:::
::::

[^1]: Some footnote text
````

into this:

````markdown
Here's a paragraph with a footnote[^1][^2].

[^1]: Some footnote text

[^2]: And here's a longer margin block-level element...
      :::{note}
      ...with a note inside.
      :::
      ::::
````

Or put another way, are there situations where somebody would want to define margin content where they don't want an explicit reference to the text?

AakashGfude:
Cool, I will try out "sidenotes via footnotes syntax" in a new PR.

choldgraf:
closing this in favor of #546

PR #542, opened 2022-03-29
Fixing figures in the margin

choldgraf:
Catches one other figure margin test case.

Thanks @da5nsy for reporting this one too - want to test if this makes it work for you?


choldgraf:
Actually, I'll just merge this in and cut a new patch release so you can test in jupyter book

PR #546, opened 2022-03-30
Sidenote using footnote syntax

AakashGfude:
Taking @mmcky and @choldgraf's comments into account. Implemented the sidenote and marginnote using footnote syntax as the first step. Changing footnote nodes to sidenote and inline element nodes in a post-transform. This does not create any new syntax to write sidenotes and uses the already available footnote syntax. To convert all footnotes to sidenotes, the config variable is - `use_sidenotes: True` under `html_theme_options`. To convert a sidenote to marginnote (i.e. not haveing superscripted numbers), use `{>}` at the start of the footnote source text (similar syntax was first proposed in https://github.com/executablebooks/sphinx-book-theme/pull/523#issuecomment-1069154993). 

**Example**:

 _Source Code:_
```
On wide screens, margin content will pop out to the side of the page and allow content underneath it to move upwards[^myref].
This allows you to provide extra information without interrupting the flow of information.

There are many other things that you can do with the Sphinx Book Theme.
Now that you've gotten a start, check out the other sections to the left to learn more about how to use it[^3].

[^myref]: Some footnote text

[^3]: And here's a longer margin block-level element...
      :::{note}
      ...with a note inside.
      :::
```

_Footnote:_

Screen Shot 2022-04-06 at 4 55 48 pm

_Sidenote:_

Screen Shot 2022-04-06 at 4 55 18 pm

When modifying the above source code with:

```
[^myref]:  {>} Some footnote text
```

The output is:

Screen Shot 2022-04-07 at 10 53 04 am

Note that sidenotes do not support any directives inside it as it is an inline element. Tufte also uses inline span elements. The reason span makes more sense than aside in my opinion is mentioned in the **31st March** description below.

**NOTE**: The functionality of clickable text/icon on smaller screens as was implemented in https://github.com/executablebooks/sphinx-book-theme/pull/523, is also present in this PR.

## 31st march

Was trying to implement sidenote leveraging the footnote code and syntax as discussed in https://github.com/executablebooks/sphinx-book-theme/pull/523 . I tried going through the docutils code, where the footnote is implemented. 
I have a few concerns which I wanted to discuss:

- the insertion of `footnote` node in the document happens during the parsing phase I think. And possibly the only place to rearrange the document to make the `footnote` node move near the `footnote-reference` is during a post-transform. 

- It was much easier to deal with `span` elements we generated here https://github.com/executablebooks/sphinx-book-theme/pull/523 for sidenotes, and which Tufte CSS uses as well, to position the sidenote. As `span` can go inside a `p` tag, and other similar tags, which will have the sidenote reference inside it. And so relative positioning is pretty straightforward. The `aside` tag on the other hand, when trying to insert inside a 'p' tag in post-transform, does not work, and it is always inserted after the `p` tag, which makes it hard to position. 

cc: @choldgraf @chrisjsewell @mmcky @rowanc1

### DIFF
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -88,7 +88,7 @@ but I'll stop here.
 (margin:sidenote)=
 ## Sidenotes and marginnotes
 
-This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/), with a UX similar to [pandoc-sidenote](https://github.com/jez/pandoc-sidenote).
+This theme has support for [Tufte-style margin / side notes](https://edwardtufte.github.io/tufte-css/), with a UX similar to [pandoc-sidenote](https://github.com/jez/pandoc-sidenote).
 
 Sidenotes are numbered, and behave like footnotes, except they live in the margin and don’t force the reader to jump their eye to the bottom of the page.
 For example, here is a sidenote[^ex].
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -88,7 +88,7 @@ but I'll stop here.
 (margin:sidenote)=
 ## Sidenotes and marginnotes
 
-This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/). The UX of which has been inspired by [pandoc-sidenote](https://github.com/jez/pandoc-sidenote)
+This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/), with a UX similar to [pandoc-sidenote](https://github.com/jez/pandoc-sidenote).
 
 Sidenotes are numbered, and behave like footnotes, except they live in the margin and don’t force the reader to jump their eye to the bottom of the page.
 For example, here is a sidenote[^ex].
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -88,7 +88,7 @@ but I'll stop here.
 (margin:sidenote)=
 ## Sidenotes and marginnotes
 
-This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/).
+This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/). The UX of which has been inspired by [pandoc-sidenote](https://github.com/jez/pandoc-sidenote)
 
 Sidenotes are numbered, and behave like footnotes, except they live in the margin and don’t force the reader to jump their eye to the bottom of the page.
 For example, here is a sidenote[^ex].
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -67,7 +67,7 @@ I could try to add a list of ideas to talk about.
 I suppose I could just keep going on forever,
 but I'll stop here.
 
--- Jo the Jovyan, *[the jupyter book docs](https://beta.jupyterbook.org)*
+-- Jo the Jovyan, *[the jupyter book docs](https://jupyterbook.org)*
 ```
 
 `````{toggle}
@@ -80,7 +80,7 @@ I could try to add a list of ideas to talk about.
 I suppose I could just keep going on forever,
 but I'll stop here.
 
--- Jo the Jovyan, *[the jupyter book docs](https://beta.jupyterbook.org)*
+-- Jo the Jovyan, *[the jupyter book docs](https://jupyterbook.org)*
 ```
 ````
 `````
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -21,30 +21,20 @@ There are a few content blocks that are unique to this theme.
 
 Here is what quotations and epigraphs look like in `sphinx-book-theme`:
 
-A quote with no attribution:
-
+```{example} Default markdown quotes
 > Here's my quote, it's pretty neat.
 > I wonder how many lines I can create with
 > a single stream-of-consciousness quote.
 > I could try to add a list of ideas to talk about.
 > I suppose I could just keep going on forever,
 > but I'll stop here.
+```
 
 Sometimes you'd like to draw more attention to a quote. To do so, use the `{epigraph}` directive.
 Below is an epigraph, click the button to the right of it to show the code that was used
 to generate it:
 
-```{epigraph}
-Here's my quote, it's pretty neat.
-I wonder how many lines I can create with
-a single stream-of-consciousness quote.
-I could try to add a list of ideas to talk about.
-I suppose I could just keep going on forever,
-but I'll stop here.
-```
-
-`````{toggle}
-````
+````{example} A basic epigraph
 ```{epigraph}
 Here's my quote, it's pretty neat.
 I wonder how many lines I can create with
@@ -54,24 +44,11 @@ I suppose I could just keep going on forever,
 but I'll stop here.
 ```
 ````
-`````
 
 You can also add an attribution to epigraphs by adding a blank line,
 followed by a line that starts with `--`. This will be renderered like so:
 
-```{epigraph}
-Here's my quote, it's pretty neat.
-I wonder how many lines I can create with
-a single stream-of-consciousness quote.
-I could try to add a list of ideas to talk about.
-I suppose I could just keep going on forever,
-but I'll stop here.
-
--- Jo the Jovyan, *[the jupyter book docs](https://jupyterbook.org)*
-```
-
-`````{toggle}
-````
+````{example} Epigraphs with attribution
 ```{epigraph}
 Here's my quote, it's pretty neat.
 I wonder how many lines I can create with
@@ -83,7 +60,6 @@ but I'll stop here.
 -- Jo the Jovyan, *[the jupyter book docs](https://jupyterbook.org)*
 ```
 ````
-`````
 
 (margin:sidenote)=
 ## Sidenotes and marginnotes
@@ -168,20 +144,22 @@ If you'd like in-line margin content with numbered references, see [](margin:sid
 
 Here's how you can use the `{margin}` directive:
 
+````{example}
+:no-container:
+:reverse:
+
 ```{margin} **Here is my margin content**
 It is pretty cool!
 ```
-
-````
-```{margin} **My margin title**
-Here is my margin content, it is pretty cool!
-```
 ````
 
-## Figure captions in the margin
+### Figure captions in the margin
 
 You can configure figures to use the margin for captions.
 Here is a figure with a caption to the right.
+
+::::{example}
+:no-container:
 
 ```{figure} images/cool.jpg
 ---
@@ -192,20 +170,7 @@ name: myfig5
 ---
 And here is my figure caption, if you look to the left, you can see that COOL is in big red letters. But you probably already noticed that, really I am just taking up space to see how the margin caption looks like when it is really long :-).
 ```
-
-And the text that produced it:
-
-````
-```{figure} images/cool.jpg
----
-width: 60%
-figclass: margin-caption
-alt: My figure text
-name: myfig5
----
-And here is my figure caption, if you look to the left, you can see that COOL is in big red letters. But you probably already noticed that, really I am just taking up space to see how the margin caption looks like when it is really long :-)
-```
-````
+::::
 
 We can reference the figure with {ref}`this reference <myfig5>`. Or a numbered reference like
 {numref}`myfig5`.
@@ -217,37 +182,26 @@ To do so, add the `margin` CSS class to any element on the page.
 Many Sphinx directives allow you to directly add classes.
 For example, here's the syntax to add a `margin` class to a `{note}` directive:
 
-:::{note}
-:class: margin
-This note will be in the margin!
-:::
+```{example}
+:no-container:
+:reverse:
 
-```
 :::{note}
 :class: margin
 This note will be in the margin!
 :::
 ```
+
 
 This works for most elements on the page, but in general this works best for "parent containers" that are the top-most element of a bundle of content.
-
-For example, we can even put a whole figure in the margin like so:
-
-
-You can also put the whole figure in the margin if you like.
+For example, you can also put the whole figure in the margin if you like.
 Here is a figure with a caption below. We'll add a note below to create
 some vertical space to see better.
 
-```{figure} images/cool.jpg
----
-figclass: margin
-alt: My figure text
-name: myfig4
----
-And here is my figure caption
-```
+````{example}
+:no-container:
+:reverse:
 
-````
 ```{figure} images/cool.jpg
 ---
 figclass: margin
@@ -262,7 +216,7 @@ We can reference the figure with {ref}`myfig4`. Or a numbered reference like
 {numref}`myfig4`.
 
 
-### Adding content to margins and sidebars
+### Content examples in the margin
 
 Margin content can include all kinds of things, such as code blocks:
 
@@ -272,7 +226,6 @@ print("here is some python")
 ```
 ````
 
-`````
 ````{margin} Code blocks in margins
 ```python
 print("here is some python")
@@ -280,16 +233,10 @@ print("here is some python")
 ````
 `````
 
-as well as admonitions and images:
+`````{example} Admonitions and images in the margin
+:no-container:
+:reverse:
 
-````{margin} **Notes in margins**
-```{note}
-Wow, a note with an image in a margin!
-![](images/cool.jpg)
-```
-````
-
-`````
 ````{margin} **Notes in margins**
 ```{note}
 Wow, a note with an image in a margin!
@@ -298,6 +245,28 @@ Wow, a note with an image in a margin!
 ````
 `````
 
+## Sidebars
+
+Sidebars exist in-line with your text, but allow the rest of the
+page to flow around them, rather than moving to the right margin.
+
+To add content sidebars, use this syntax:
+
+`````{example}
+:no-container:
+
+````{sidebar} **My sidebar title**
+```{note}
+Here is my sidebar content, it is pretty cool!
+```
+![](images/cool.jpg)
+````
+`````
+
+Note how the content wraps around the sidebar to the right.
+However, the sidebar text will still be in line with your content. There are
+certain kinds of elements, such as "note" blocks and code cells, that may
+clash with your sidebar. If this happens, try using a `{margin}` instead.
 
 ## Full-width content
 
@@ -306,47 +275,17 @@ the rest of your book's content. To add full-width content to your page, add the
 class `full-width` to any of the elements in your documentation. For example, you can
 add a `full-width` tag to a note element like this:
 
-````
-```{note}
-:class: full-width
-This content will be full-width
-```
-````
-
-This code results in the following output:
+````{example}
+:no-container:
 
 ```{note}
 :class: full-width
 This content will be full-width
 ```
+````
 
 ```{admonition} A note for ipynb users
 If you are using a Jupyter Notebook as inputs to your documentation using the
 [MyST-NB extension](https://myst-nb.readthedocs.io/en/latest/), you can trigger
 this behavior with a code cell by adding a `full-width` tag to the cell.
 ```
-
-
-## Sidebars
-
-Content sidebars exist in-line with your text, but allow the rest of the
-page to flow around them, rather than moving to the right margin.
-To add content sidebars, use this syntax:
-
-````{sidebar} **My sidebar title**
-```{note}
-Here is my sidebar content, it is pretty cool!
-```
-![](images/cool.jpg)
-````
-
-Note how the content wraps around the sidebar to the right.
-However, the sidebar text will still be in line with your content. There are
-certain kinds of elements, such as "note" blocks and code cells, that may
-clash with your sidebar. If this happens, try using a `{margin}` instead.
-
-````
-```{sidebar} **My sidebar title**
-Here is my sidebar content, it is pretty cool!
-```
-````
--- a/docs/content-blocks.md
+++ b/docs/content-blocks.md
@@ -85,31 +85,86 @@ but I'll stop here.
 ````
 `````
 
-## Sidebars
+(margin:sidenote)=
+## Sidenotes and marginnotes
 
-There are two different kinds of sidebar-like content in `sphinx-book-theme`,
-typical `{sidebar}` directives, as well as a theme-specific `{margin}` directive.
-This section covers both. Both allow you to place extra content
-separately from your main content.
+This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/).
 
-```{tip}
-Sidebar content will generally overlap with the white space where your site's
-table of contents lives. When the reader scrolls sidebar content into view, the
-right TOC should hide itself automatically.
+Sidenotes are numbered, and behave like footnotes, except they live in the margin and don’t force the reader to jump their eye to the bottom of the page.
+For example, here is a sidenote[^ex].
+On narrow screens, sidenotes are hidden until a user clicks the number.
+If you're on a mobile device, try clicking the sidenote number above.
+
+[^ex]: Here's my sidenote text!
+
+    On narrow screens, this text won't show up unless you click the superscript number!
+
+Marginnotes are not numbered, but behave the same way as sidenotes.
+On mobile devices a symbol will be displayed that will display the marginnote when clicked[^exmn].
+For example, there's a marginnote in the previous sentence, and you should see a symbol show to display it on mobile screens.
+
+[^exmn]: {-} This is a margin note. Notice there isn’t a number preceding the note.
+
+:::{seealso}
+Sidenotes and marginnotes are inline content - you cannot use block-level content inside of these notes.
+If you'd like to use block-level content in the margins, see [](margin:block).
+:::
+
+### Activate sidenotes and marginnotes
+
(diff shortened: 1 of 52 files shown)


mmcky:
@AakashGfude I have been taking a look at this PR in addition to #523 and from a `developer` perspective it seems to me that it would make sense to:

1. Develop the concept of a `sidenote` as a feature with supporting `visit` and `depart` methods for `html` and `latex`
2. Add a global option `footnote_as_sidenote: [True/False]` which enables a `sphinx.transform` to move `footnote` objects (visually) to `sidenote` objects in the `AST` to be parsed by `sidenote` writers. 

To me this flow seems more logical rater than overloading the `footnote` objects to create `sidenote` style of output. Some of the benefits of this approach are improvements made to `sidenote` feature are automatically gained if we use a `transform` to move `footnotes` to `sidenote` nodes in the AST.

AakashGfude:
Thanks, @mmcky, that sounds like a reasonable way to go. I will approach it that way and see how it feels.

AakashGfude:
made changes, and updated the top description.

choldgraf (commented):
This is looking nice! Much cleaner than having lots of new roles/directives, I think. A few quick thoughts:

- I pushed a commit that added the extra docs from your other PR, and updated them a bit for this new syntax.
- I updated the "marginnote" prefix to `{-}` instead of `{>}` to be in-line with what the pandoc-sidenotes extension uses.
- I've added some example usage in our reference docs as well so we can show off the behavior there.
- I noticed that having multiple marginnotes in the same sentence does not work on mobile screens, I think it's because the `for` is pointing to the same thing in both cases, can you fix that?
- I think that it'd be good to have some visual cue that the numbers/symbols are "clickable" on mobile - maybe they can still be blue? Or some other way to suggest they can be clicked?

AakashGfude:
Thank you so much @choldgraf, for the additions, very useful. 

>     * I noticed that having multiple margin notes in the same sentence does not work on mobile screens, I think it's because the `for` is pointing to the same thing in both cases, can you fix that?

Have pushed the fix now.

>     * I think that it'd be good to have some visual cue that the numbers/symbols are "clickable" on mobile - maybe they can still be blue? Or some other way to suggest they can be clicked?

Yes, I think the original color (blue) seems like a good option. As the user is already used to assuming that it is a clickable reference? Have pushed the change. Let's see how it feels.

choldgraf:
I think the current implementation is pretty good to me. The behavior is nearly identical to how [pandoc-sidenotes](https://github.com/jez/pandoc-sidenote) works, so we're not reinventing the wheel from a UX perspective or anything.

One thing I noticed: If you have a marginnote in between two sidenotes, it still takes up a "number" even though it is not actually numbered (so the sidenotes will jump from 1 to 3 even though there is no 2). Is that easy to fix? If not it's not a blocker on this PR in my opinion.

src/sphinx_book_theme/nodes.py
```diff
@@ -0,0 +1,48 @@
...
+        add_node(
+            cls,
+            override=True,
+            html=(visit_SideNoteNode, depart_SideNoteNode),
+            latex=(visit_SideNoteNode, depart_SideNoteNode),
+        )
+
+
+def visit_SideNoteNode(self, node):
+    if isinstance(self, LaTeXTranslator):
```
chrisjsewell, line 25: The post transform should be set to only run on the html format, then you can get rid of this

chrisjsewell (changes requested):

src/sphinx_book_theme/nodes.py
AakashGfude: sounds good. The way to handle it in latex should probably be in `sphinx-jupyterbook-latex`? At present, the pdf just ignores margin texts. 
Unless we decide to move this functionality out of `sphinx-book-theme`

chrisjsewell: Yep exactly 👍, whilst it is only in sphinx-book-theme, it should only affect html

AakashGfude:
> One thing I noticed: If you have a marginnote in between two sidenotes, it still takes up a "number" even though it is not actually numbered (so the sidenotes will jump from 1 to 3 even though there is no 2). Is that easy to fix? If not it's not a blocker on this PR in my opinion.

ooh yeah, it's because we are converting footnotes to sidenotes/margin notes and using the numbering of footnotes. margin notes were also footnotes and had a number.

src/sphinx_book_theme/nodes.py
AakashGfude: cool, have added check for `standaloneHTMLBuilder`.

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,52 @@
...
+from .nodes import SideNoteNode
+from sphinx.builders.html import StandaloneHTMLBuilder
+
+
+class HandleFootnoteTransform(SphinxPostTransform):
+
+    default_priority = 1
+
+    def apply(self, **kwargs: Any) -> None:
+        if isinstance(self.env.app.builder, StandaloneHTMLBuilder):
```
chrisjsewell, line 13: No you just add `formats = ("html",)` (have a look at the sphinx post transform documentation)

chrisjsewell (changes requested):

src/sphinx_book_theme/_transforms.py
AakashGfude: pretty neat. Thanks, have used that now.

AakashGfude:
@choldgraf This looks ready to go 🚀

mmcky:
@AakashGfude this is looking great. Thanks for working on this. 

Will this get "inherited" by any downstream themes such as `quantecon-book-theme`? 

> Apart from mentioning pandoc-sidenote. Thought you would be a better person to put it in the docs.

Do you mind giving this a shot to complete this PR. I can pitch in with some suggestions if needed -- feel free to slack/zoom.

AakashGfude:
> Will this get "inherited" by any downstream themes such as `quantecon-book-theme`?
 yes, it will. 

> Do you mind giving this a shot to complete this PR. I can pitch in with some suggestions if needed -- feel free to slack/zoom.

I have done a futile attempt here https://github.com/executablebooks/sphinx-book-theme/pull/546/commits/8cc71d9e1be5e5c6d1ba0cef2f82a9dd84c450e0

docs/content-blocks.md
```diff
@@ -85,32 +85,87 @@ but I'll stop here.
 ````
 `````
 
-## Sidebars
+(margin:sidenote)=
+## Sidenotes and marginnotes
+
+This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/). The UX of which has been inspired by [pandoc-sidenote](https://github.com/jez/pandoc-sidenote)
```
choldgraf, line 91:
```suggestion
This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/), with a UX similar to [pandoc-sidenote](https://github.com/jez/pandoc-sidenote).
```

AakashGfude:
@choldgraf should we merge this?

docs/content-blocks.md
```diff
@@ -85,32 +85,87 @@ but I'll stop here.
 ````
 `````
 
-## Sidebars
+(margin:sidenote)=
+## Sidenotes and marginnotes
+
+This theme has support for [Tufte-stype margin / side notes](https://edwardtufte.github.io/tufte-css/), with a UX similar to [pandoc-sidenote](https://github.com/jez/pandoc-sidenote).
```
choldgraf, line 91:
```suggestion
This theme has support for [Tufte-style margin / side notes](https://edwardtufte.github.io/tufte-css/), with a UX similar to [pandoc-sidenote](https://github.com/jez/pandoc-sidenote).
```

docs/content-blocks.md
```diff
@@ -85,32 +85,87 @@ but I'll stop here.
...
+
+    On narrow screens, this text won't show up unless you click the superscript number!
+
+Marginnotes are not numbered, but behave the same way as sidenotes.
+On mobile devices a symbol will be displayed that will display the marginnote when clicked[^exmn].
+For example, there's a marginnote in the previous sentence, and you should see a symbol show to display it on mobile screens.
+
+[^exmn]: {-} This is a margin note. Notice there isn’t a number preceding the note.
+
+:::{seealso}
```
choldgraf, line 108: I noticed that sidenotes and marginnotes don't work if you put them inside of admonitions. They get placed _inside_ the admonition rather than in the margin. Is this a relatively easy fix? If not then we could add a note that says this isn't possible and try fixing it in a follow-up PR.

docs/reference/special-theme-elements.md
```diff
@@ -271,6 +270,17 @@ alt: My figure text
...
+
+[^sn1]: Test sidenote 1.
+[^sn2]: Test sidenote 2.
+
+Here's a sentence[^mn1] with multiple marginnotes[^mn2].
+
+[^mn1]: {-} Test marginnote 1.
+[^mn2]: {-} Test marginnote 2.
 
 ## Nested admonitions
```
choldgraf, line 285:
```suggestion
[^mn1]: {-} Test marginnote 1.
[^mn2]: {-} Test marginnote 2.


Sidenotes inside of admonitions should behave the same:

:::{note}
An admonition with a sidenote defined in the admonition[^snam1] and another defined outside of the admonition [^snam2].

[^snam1]: Sidenote defined in the admonition.

:::

[^snam2]: Sidenote defined outside the admonition.


## Nested admonitions
```

choldgraf: Here's the text I tried to test the buggy behavior when you define sidenotes in an admonition

src/sphinx_book_theme/assets/styles/content/_margin.scss
```diff
@@ -38,6 +38,48 @@ $content-fullwidth-width: percentage(100% / $content-max-width);
   }
 }
 
+label.margin-toggle {
```
choldgraf, line 41: Can we add a `user-select: none` somewhere for the superscript number?

https://css-tricks.com/almanac/properties/u/user-select/

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class HandleFootnoteTransform(SphinxPostTransform):
+
+    default_priority = 1
+    formats = ("html",)
+
+    def run(self, **kwargs: Any) -> None:
+        theme_options = self.env.config.html_theme_options
+        if theme_options.get("use_sidenotes", False) is True:
```
choldgraf, line 14: To remove an extra indendation, could this instead be:

```python
if theme_options.get("use_sidenotes", False) is False:
    return None
...and then continue the logic w/o indentation here
```

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,51 @@
...
+                        == node.attributes["ids"][0]
+                    ):
+                        parent = ftnode.parent
+                        text = ftnode.children[1].astext()
+
+                        sidenote = SideNoteNode()
+                        para = docutil_nodes.inline()
+                        label = ftnode.children[0].astext()
+                        if text.startswith("{-}"):
+                            # marginnotes
```
choldgraf, line 31:
```suggestion
                            # marginnotes will have content starting with {-}
                            # remove the number so it doesn't show
```

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,51 @@
...
+                            para.attributes["classes"].append("marginnote")
+                            para.append(docutil_nodes.Text(text.replace("{-}", "")))
+
+                            sidenote.attributes["names"].append(
+                                f"marginnote-role-{label}"
+                            )
+                        else:
+                            # sidenotes
```
choldgraf, line 39:
```suggestion
                            # sidenotes are the default behavior if no {-}
                            # in this case we keep the number
```

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,51 @@
...
+                            superscript = docutil_nodes.superscript("", label)
+                            para.attributes["classes"].append("sidenote")
+                            para.extend([superscript, docutil_nodes.Text(text)])
+
+                            sidenote.attributes["names"].append(
+                                f"sidenote-role-{label}"
+                            )
+                            sidenote.append(superscript)
+                        node.replace_self([sidenote, para])
```
choldgraf, line 48: I guess somewhere around here we'd need to look for a parent block and place the `para` in the parent instead of right next to the sidenote? Feels like this might be too much complexity to worry about this edge-case...

choldgraf (approved):
A few more quick comments, but I think we can try to resolve and merge quickly. If anything seems like it'll take a while, let's just merge it and fix in a later PR. I'm +1 to merge after comments addressed

choldgraf:
Friendly ping here @AakashGfude - let me know if you don't have time for another cycle here, or if you'd like to give this another quick push so we can merge

AakashGfude:
> Friendly ping here @AakashGfude - let me know if you don't have time for another cycle here, or if you'd like to give this another quick push so we can merge

Hi, @choldgraf will review this tomorrow, was traveling for a few days.

choldgraf:
@AakashGfude no worries - I just wanted to see if you still had the bandwidth for another cycle or if you wanted to spot-check these things in future PRs. I think anything that turns out to be complex doesn't need to be a blocker on this PR, but figured I'd give a final round of feedback in case it helps :-)

AakashGfude:
> @AakashGfude no worries - I just wanted to see if you still had the bandwidth for another cycle or if you wanted to spot-check these things in future PRs. I think anything that turns out to be complex doesn't need to be a blocker on this PR, but figured I'd give a final round of feedback in case it helps :-)

Yes for sure, thanks for taking out the time to suggest further helpful improvements. Really appreciate it :)

choldgraf:
I think for the "nested footnotes" bug, one hacky (but maybe reasonable?) way to solve this would be to do something like:

- Move the footnote content next to the footnote reference, like we do now
- Check whether the *parent* of this location is the document. If not, then the footnote content to *just before* the parent node.
- Repeat this process until the footnote content's parent is now the document itself

src/sphinx_book_theme/assets/styles/content/_margin.scss
AakashGfude: Added here https://github.com/executablebooks/sphinx-book-theme/pull/546/commits/75d40af99d42133cd7d5c9fd70a14518bf2eea20

src/sphinx_book_theme/_transforms.py
AakashGfude: sure thing. Done in https://github.com/executablebooks/sphinx-book-theme/pull/546/commits/bb485b6ccbce37083a0df073f1a174e380b3a756

AakashGfude:
Thanks for the commits and the suggestions @choldgraf.
And also for putting forward your suggestion for the algo,i.e., nested sidenotes/margin notes. Have implemented a basic one in https://github.com/executablebooks/sphinx-book-theme/pull/546/commits/ac9dcabf10015fb7feea6b159d9aaf8879efa81e. Let me know if you can think of edge cases. I will write a test case for it tomorrow morning, and check for any improvements.

choldgraf:
It looks good! One UX bug I noticed is that on narrow screens for a side/marginnote inside an admonition, clicking it doesn't seem to have an effect

AakashGfude:
> It looks good! One UX bug I noticed is that on narrow screens for a side/marginnote inside an admonition, clicking it doesn't seem to have an effect.

Yeah, this is a tricky one, as we have placed the text outside the admonition. Will have to think of something.

AakashGfude:
> It looks good! One UX bug I noticed is that on narrow screens for a side/marginnote inside an admonition, clicking it doesn't seem to have an effect.

@choldgraf has resolved it, by keeping two copies of the `side/marginnote` text. One outside admonition as discussed. And one in the original place which remains hidden when on a desktop. When using mobile, the hidden one is visible. 


Screen Shot 2022-05-13 at 11 33 53 am

What do you think?

src/sphinx_book_theme/assets/styles/base/_base.scss
```diff
@@ -10,6 +10,10 @@
   left: 0;
 }
 
+.d-n {
```
choldgraf, line 13: Can you just use `d-none` for now since we are inheriting bootstrap's classes anyway?

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,70 @@
...
+from docutils import nodes as docutil_nodes
+from sphinx import addnodes as sphinx_nodes
+from .nodes import SideNoteNode
+import copy
+
+
+class HandleFootnoteTransform(SphinxPostTransform):
+
+    default_priority = 1
+    formats = ("html",)
```
choldgraf, line 12:
```suggestion
class HandleFootnoteTransform(SphinxPostTransform):    
    """Transform footnotes into side/marginnotes."""    
    default_priority = 1
    formats = ("html",)
```

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def run(self, **kwargs: Any) -> None:
+        theme_options = self.env.config.html_theme_options
+        if theme_options.get("use_sidenotes", False) is False:
+            return None
+        for node in self.document.traverse(docutil_nodes.footnote_reference):
```
choldgraf, line 18:
```suggestion
        # Cycle through footnote references, and move their content next to the reference
        # This lets us display the reference in the margin, or just below on narrow screens.
        for node in self.document.traverse(docutil_nodes.footnote_reference):
```

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,70 @@
...
+            parent = None
+            for ftnode in self.document.traverse(docutil_nodes.footnote):
```
choldgraf, line 20:
```suggestion
            # Each footnote reference should have a single node it points to via `ids`
            for ftnode in self.document.traverse(docutil_nodes.footnote):
```

src/sphinx_book_theme/_transforms.py
choldgraf, line 18: Can you use a more descriptive word than `node`? How about `ref_node` and below `foot_node`.

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,70 @@
...
+                # matching the footnote reference with footnote
+                if (
+                    len(ftnode.attributes["backrefs"])
+                    and ftnode.attributes["backrefs"][0] == node.attributes["ids"][0]
+                ):
+                    parent = ftnode.parent
+                    text = ftnode.children[1].astext()
```
choldgraf, line 27: Explain why we do `children[1]`. What is the first element of `children`? What is the element we want to grab? This can be helpful in case docutils changes this ordering in the future

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,70 @@
...
+
+                    sidenote = SideNoteNode()
+                    para = docutil_nodes.inline()
+                    label = ftnode.children[0].astext()
+
+                    if text.startswith("{-}"):
+                        # marginnotes will have content starting with {-}
+                        # remove the number so it doesn't show
+                        para.attributes["classes"].append("marginnote")
+                        para.append(docutil_nodes.Text(text.replace("{-}", "")))
```
choldgraf, line 37: Should we do a `.strip()` here as well since there will be a space after the `{-}`?

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,70 @@
...
+                        # sidenotes are the default behavior if no {-}
+                        # in this case we keep the number
+                        superscript = docutil_nodes.superscript("", label)
+                        para.attributes["classes"].append("sidenote")
+                        para.extend([superscript, docutil_nodes.Text(text)])
+
+                        sidenote.attributes["names"].append(f"sidenote-role-{label}")
+                        sidenote.append(superscript)
+
+                    # for nested footnotes/marginnotes
```
choldgraf, line 50:
```suggestion
                    # If the reference is nested (e.g. in an admonition), duplicate the content node
                    # And place it just before the parent container, so it works w/ margin
                    # Only show one or another depending on screen width.
```

src/sphinx_book_theme/_transforms.py
```diff
@@ -0,0 +1,70 @@
...
+                        if not isinstance(
+                            node_parent,
+                            (docutil_nodes.paragraph, docutil_nodes.footnote),
+                        ):
+                            node_parent.replace_self([para, node_parent])
+                            para_dup.attributes["classes"].append("d-n")
+                            break
+                        node_parent = node_parent.parent
+
+                    node.replace_self([sidenote, para_dup])
```
choldgraf, line 67: Rather than `para_dup` can we be more specific? Like `para_nested` and `para_margin`?

src/sphinx_book_theme/assets/styles/content/_margin.scss
```diff
@@ -38,6 +37,54 @@ $content-fullwidth-width: percentage(100% / $content-max-width);
...
```
choldgraf, line 40:
```suggestion
// Sidenotes and marginnotes
label.margin-toggle {
```

src/sphinx_book_theme/nodes.py
```diff
@@ -0,0 +1,36 @@
+from docutils import nodes
+from sphinx.application import Sphinx
+from typing import Any, cast
+
+
+class SideNoteNode(nodes.Element):
+    """A node that will not be rendered."""
```
choldgraf, line 7: This doesn't sound like the correct docstring

choldgraf (commented):
Thanks for these updates @AakashGfude - I've added a few suggestions and comments in there, mostly around trying to make it explicit what the code is doing so that it is more maintainable. Let me know what you think!

src/sphinx_book_theme/assets/styles/base/_base.scss
AakashGfude: sure thing

src/sphinx_book_theme/_transforms.py
AakashGfude: We can I guess. But without `strip()` it seems to be stripping spaces as well, no matter how much space we give.

src/sphinx_book_theme/_transforms.py
AakashGfude: sounds good.

src/sphinx_book_theme/nodes.py
AakashGfude: Oops.

src/sphinx_book_theme/_transforms.py
AakashGfude: @choldgraf The variable itself is just a duplicate of para. So, I thought the name should just signify that without adding any extra connotations. Else I can use `para_nested` probably?

src/sphinx_book_theme/nodes.py
AakashGfude: done in [a87245f](https://github.com/executablebooks/sphinx-book-theme/pull/546/commits/a87245f0706bbf8c070a78a6329783289cc7f1e1)

src/sphinx_book_theme/_transforms.py
AakashGfude: done in [f5efddf](https://github.com/executablebooks/sphinx-book-theme/pull/546/commits/f5efddf4772fa3d2f39b3755ecabe5d228b90e54)

src/sphinx_book_theme/_transforms.py
choldgraf: fair enough - maybe this is a sphinx parsing thing under the hood

src/sphinx_book_theme/assets/styles/base/_base.scss
```diff
@@ -10,6 +10,10 @@
   left: 0;
 }
 
+.d-none {
+  display: none;
+}
+
```
choldgraf, line 16:
```suggestion
```

My point was that I don't think we need this at all, because bootstrap should already give it to us

AakashGfude: aaah yeah. ofcourse.

choldgraf:
OK I think I fixed up the docs conflict, wanna take a look and let me know if it reads OK to you?

Other than that, anything else you'd like to tackle here?

AakashGfude:
@choldgraf the doc looks good to me. Thank you! 
I could not use the `d-none` class instead of `d-n` like you mentioned here https://github.com/executablebooks/sphinx-book-theme/pull/546#discussion_r872050295. As `d-none` class in pydata-sphinx-theme has `display:none !important` in it, which is preventing it from being overridden. 
 
Else I think rest is good.

src/sphinx_book_theme/assets/styles/base/_base.scss
```diff
@@ -23,6 +23,10 @@
   width: 1px !important;
 }
 
+.d-n {
```
choldgraf, line 26:
```suggestion
// We need to define our own display-none class since bootstrap uses !important and we want to be able to over-ride
.d-n {
```

choldgraf (approved):
OK this looks good to me - let's merge it in and see how folks like using it, and we can iterate from there!

choldgraf:
Thanks very much @AakashGfude for this addition!

AakashGfude:
Thank you, @choldgraf.

PR #556, opened 2022-04-27
FIX: Escape colon in JupyterHub link to repo

matthew-brett:
I've run into a problem using The Littlest JupyterHub with current
`sphinx-book-theme`. At the moment the generated URLs for JupyterHub are
of form:

```
<hub-url>/hub/user-redirect/git-pull?repo=https://<repo>&urlpath=tree/<nb_fname>
```

These work fine on a modern Kubernetes JupyterHub, I guess because of
a recent version of JupyterHub responding to the query.

But, on current The Littlest JupyterHub, still on JupyterHub 1.5.0, this
causes Nbgitpuller to stall if the user is not already logged in, with
log messages of form:

```
Disallowing redirect outside JupyterHub:
'/hub/user-redirect/git-pull?repo=https://github.com/<repo>&urlpath=tree/<nb_fname>'.
```

Once logged in, retrying the same link works correctly.

This can be fixed by escaping the `:` in the Nbgitpuller URL with a
`%3A`, in this PR.

### DIFF
--- a/src/sphinx_book_theme/header_buttons/launch.py
+++ b/src/sphinx_book_theme/header_buttons/launch.py
@@ -118,9 +118,10 @@ def add_launch_buttons(
         )
 
     if jupyterhub_url:
+        repo_esc = repo_url.replace(':', '%3A')
         url = (
             f"{jupyterhub_url}/hub/user-redirect/git-pull?"
-            f"repo={repo_url}&urlpath={ui_pre}/{repo}/{path_rel_repo}&branch={branch}"
+            f"repo={repo_esc}&urlpath={ui_pre}/{repo}/{path_rel_repo}&branch={branch}"
         )
         launch_buttons_list.append(
             {
--- a/src/sphinx_book_theme/header_buttons/launch.py
+++ b/src/sphinx_book_theme/header_buttons/launch.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 from typing import Any, Dict, Optional
+from urllib.parse import urlencode
 
 from docutils.nodes import document
 from sphinx.application import Sphinx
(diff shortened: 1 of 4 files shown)


choldgraf (commented):
This seems good to me - can you confirm that the link still works on a "modern" JupyterHub as well? Also it looks like you need to update your branch to the latest `master`

matthew-brett:
I've rebased.

I tested a modern Kubernetes setup by replacing the `:` in the generated URL with `%3A` - it worked as expected.

I see that a few tests will need fixing if this is the right way to go.

choldgraf:
It's not realistic for us to add tests for JupyterHubs specifically, so I think it's fine to use this PR as confirmation that "it works" and just get the tests into a working state (my guess is the regression tests will need fixing - to update those we can just replace the relevant launch links w/ the escaped version)

src/sphinx_book_theme/header_buttons/launch.py
```diff
@@ -118,9 +118,10 @@ def add_launch_buttons(
         )
 
     if jupyterhub_url:
+        repo_esc = repo_url.replace(":", "%3A")
```
minrk, line 121: A more general replacement is

```python
from urllib.parse import quote
repo_esc = quote(repo_url)
```

This will happen automatically if you use url-building functions to build the query, rather than by hand:

```python
from urllib.parse import urlencode
url_params = urlencode(dict(repo=repo_url, urlpath=f"{ui_pre}/{repo}/{path_rel_repo}", branch=branch), safe="/")
url = f"{jupyterhub_url}/hub/user-redirect/git-pull?{url_params}"

matthew-brett: Aha - I was missing the `safe="/"` kwarg - that makes all the difference to readability.

minrk: it does! I don't know why `safe="/"` is the default for `urllib.parse.quote`, but not `urllib.parse.urlencode`, but it should be fine for our purposes here.

choldgraf: @matthew-brett wanna give this a shot so that our implementation is a bit less special-casey ?

matthew-brett: Have done ...

choldgraf (approved):
😍 I love it, thanks for the follow-up, and to @minrk for helpful guidance!

Issue #162, opened 2020-08-07
Custom Title/Landing Page


chrisjsewell:
Most books that I've seen have a front cover 😉 

I think it would be great to have a front page which has a very distinctly different layout to the rest of the pages.
It might even be better if (somehow) this page is not even parsed from a standard rST/Markdown page,
but is related to a bunch of variables, and a separate HTML template (similar to [LaTeX title pages](https://www.overleaf.com/learn/latex/How_to_Write_a_Thesis_in_LaTeX_(Part_5):_Customising_Your_Title_Page_and_Abstract)).
e.g. something like you specify the variables:

```yaml
title: This is my title
author: My Name
abstract: |
    This is my abstract
```

which get parsed into a standard or custom frontpage template:

```html
<h1>{{ title }}</h1>
<p>by {{ author }}</p>
```

(but obviously more fancy like https://dev.to/davidepacilio/40-free-html-landing-page-templates-3gfp)

I don't know if this needs to be something implemented upstream (like https://github.com/pandas-dev/pydata-sphinx-theme/issues/146) or we can do here.

choldgraf:
I agree something like this would be quite useful. I guess my question is whether we should special-case this at the level of HTML / templating / etc, or if we should ensure that we provide the right Markdown syntax elements + documentation to customize them so that people can accomplish this on their own.

I suspect that the latter would be easier to maintain if we were able to get where we wanted. And Sphinx already treats the `master_doc` page as "special" and not integrated w/ the rest of the book structure. 

For certain page elements, like the title / author / etc, I think we could definitely special-case some page / site variables etc. We may want to explore that "title/author" pattern for other pages too.

btw, I like that our feature voting thing is now encouraging people to +1 their own issues 😅

choldgraf:
I looked into this a little bit more and it sounds like it should be pretty doable with a custom template + this config in Sphinx:

```python
html_additional_pages = {
    'index': 'landingpage.html',
}
```

That'll render the Jinja template `landingpage.html` to `index.html` in the built site, and we have access to all the site variables etc.

I think we'd need to over-ride Jupyter Book's current behavior regarding the `index.html` in this case (which is to add a re-direct to the first page of the book). But other than that I bet it'd be straightforward for us to bundle a nice landing page template w/ Jupyter Book and then offer a flag to turn it on.

mmcky:
[...]

PR #557, opened 2022-05-01
Prototype landing page edits

choldgraf:
This is a prototype to play around with better "landing page" style structure, using page-level metadata. Here are the major changes:

- **Adds page-level metadata**. This has a structure like the following (meant to be put in the markdown file's YAML metadata):
  
  ```yaml
  theme:
    collapse_sidebar: true
    full_width: true
  ```
  
  If that is in the page metadata it will either collapse the sidebar by default, or make the page full-width by default.
  In addition, there are now global configs for this as well.
- **Updates the padding on full-width + collapsed_sidebar**. If the article is set as full-width, and the sidebar is collapsed, then we add extra padding and center the content, so that it looks more like a typical "landing page". This should hopefully make it more suitable 

ref: #162 (I think this is a step in that direction but may not totally address it)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,5 +1,8 @@
 ---
 title: The Sphinx Book Theme
+theme:
+  collapse_sidebar: true
+  full_width: true
 ---
 
 ::::{grid}
--- a/docs/customize/index.md
+++ b/docs/customize/index.md
@@ -9,6 +9,8 @@ The following options are available via `html_theme_options`
 ```{list-table}
 :widths: 10 5 40
 :header-rows: 1
+:class: full-width
+
 * - Key
   - Type
   - Description
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ doc = [
     "numpy",
     "matplotlib",
     "numpydoc",
-    "myst-nb~=0.13",
+    "myst-nb~=0.13.2",
     "nbclient",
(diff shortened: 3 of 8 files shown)


docs/index.md
```diff
@@ -1,5 +1,8 @@
 ---
 title: The Sphinx Book Theme
+theme:
```
chrisjsewell, line 3:
```suggestion
theme_html:
```

(or `html_theme`)

this makes the purpose clearer, and would allow for e.g. the future addition of `theme_latex` etc

choldgraf: what do you think about nested?

```
theme:
    html:
```

choldgraf: ah I just realized that this might be weird to parse in Sphinx though, since the nested levels are stringified right?

choldgraf: My feeling is that we don't want to constrain MyST (or Jupyter Book) on what is possible with rST. I'd be OK with telling users "if you want nested metadata in the header, then your content should be MyST". Any objection to that?

I have a slight preference for:

```yaml
theme:
    html:
        foo: bar
    latex:
		foo: bar
```

over

```yaml
theme_html:
	foo: bar
theme_latex:
	foo: bar
```

but not a strong opinion, what do you think?

chrisjsewell: yeh, no strong preference either

PR #560, opened 2022-05-09
MAINT: Fix myst-nb pinning

choldgraf:
This fixes the myst-nb pinning in our docs and test suite, since a new release was creating problems and these weren't properly pinned.


choldgraf:
gonna merge this one since it fixes our failing tests (assuming our tests do pass here!)

PR #566, opened 2022-06-01
BUG: templates_path in config was overriding value set in extension

AakashGfude:
We update the `templates_path` config variable here https://github.com/executablebooks/sphinx-book-theme/blob/master/src/sphinx_book_theme/__init__.py#L198, to allow sidebars to be found in Jinja templates. 

But if we set a value to `templates_path` in `_config.yml`, then sphinx overrides any previous value set to it, as can be seen from this function https://github.com/sphinx-doc/sphinx/blob/5.x/sphinx/application.py#L246.  Resulting in issues like https://github.com/executablebooks/jupyter-book/issues/1660. 

This PR does the updation of the `templates_path` in `config-inited` event instead, which is emitted after the sphinx `config.init_values()` function.

fixes https://github.com/executablebooks/jupyter-book/issues/1660


mmcky:
thanks for working on this @AakashGfude. Ping me once the tests are passing and I'll review this for you.

AakashGfude:
@mmcky the tests are passing now. Have also updated the link to the sphinx-theme-builder filesystem layout.

mmcky (dismissed):
thanks @AakashGfude for fixing this bug. I had come across once as well.

mmcky:
@AakashGfude do you think we need a test case for this?

AakashGfude:
> @AakashGfude do you think we need a test case for this?

@mmcky I think that's a good idea. I will create or either modify an existing test case, to test this.

src/sphinx_book_theme/__init__.py
```diff
@@ -164,6 +164,12 @@ def run(self):
         return nodes
 
 
+def update_general_config(app):
+    theme_dir = get_html_theme_path()
+    # Update templates for sidebar
+    app.env.config.templates_path.append(os.path.join(theme_dir, "components"))
```
choldgraf, line 170: Can you add a comment explaining why this is needed given that we already add the components at the link below?

https://github.com/executablebooks/sphinx-book-theme/pull/566/files#diff-323e2d703544f60dac082d2b47d8b6a2d8e1594eb3b8f49d569f48f768379252R205

Otherwise it will confuse people why we do this twice and not just once...

AakashGfude: sure thing @choldgraf, I will add the explanation.

AakashGfude:
@mmcky and @choldgraf this issue is apparently only with jupyter-book builds. As jupyter-book calls https://github.com/executablebooks/jupyter-book/blob/master/jupyter_book/sphinx.py#L7 , which is not in the case of SphinxTestApp. Also, the error is not getting triggered with `sphinx-build`, the reason for which I am not fully aware. 

But, I had to add `jupyter-book` as a test dependency and create a jb project to test this out. Which does seem a bit too much to test this one thing.

mmcky:
@AakashGfude a good reason to write a test. That is curious. 

> Also, the error is not getting triggered with sphinx-build, the reason for which I am not fully aware.

Thanks for the update. Could it be that the `_templates` are being overwritten by `jupyter-book` cli when it compiles `sphinx-config`? Or is this condition only met when trying to load the sidebars etc. as you've identified and hence should be fixed here? If you run as a myst-nb / sphinx project the issue doesn't show up.

**Note:** this is useful when running a project without using `jb` https://jupyterbook.org/en/stable/reference/cli.html?highlight=sphinx#jupyter-book-config-sphinx

mmcky:
@AakashGfude so it doesn't seem clear to me if the fix to this templates path issue should be contained in this repository or the `jupyter-book` repository (i.e. should we be handling confoverides in a different way ). 

I just don't want to introduce circular testing dependences as it can make releases really tricky if this depends on `jupyter-book` and `jupyter-book` depends on this. If we feel it is important to test then perhaps we can place a test in `jupyter-book` to test the `_templates` issue for `sphinx-book-theme`? Is this going to be an issue for any new theme we make?

AakashGfude:
@mmcky you are right. I think we should reconsider how we are reading config in `jupyter-book`.

For instance, `sphinx-build` passes configuration to the sphinx app, using the confdir variable, which then reads the file https://github.com/sphinx-doc/sphinx/blob/5.x/sphinx/application.py#L202. But `jupyter-book` passes all configurations using confoverrides variable. Naturally, the variables will get overridden from their default values. Which we don't always want (like in the discussed case.).

AakashGfude:
> @AakashGfude so it doesn't seem clear to me if the fix to this templates path issue should be contained in this repository or the `jupyter-book` repository (i.e. should we be handling confoverides in a different way ).

I can think of only two ways we can solve this in jupyter-book. Either remove `templates_path` from confoverrides in `jupyter-book` or convert the `_config.yml` file to conf.py file and saves it in the confdir 😳. The first is probably more sensible for now.

choldgraf:
@AakashGfude where are we on this one?

AakashGfude:
> @AakashGfude where are we on this one? Are you blocked on anything? Do we need to resolve questions before moving forward?

@choldgraf This PR itself is complete I think, except for the questionable move to test this (which cannot be done without introducing jb here). And we reached a consensus that this is not the right place to fix this.. And had suggested a fix in jupyter-book .

>  Either remove `templates_path` from confoverrides in `jupyter-book`

AakashGfude:
Actually @choldgraf, there is no easy way in `jupyter-book` to achieve this without some architectural changes. I think this might be a good enough place for this for now.

choldgraf (approved):
This looks good to me - do you know what circumstances "config-inited" is not called? I thought it was always triggered but I guess not. Either way let's merge this in to fix this bug!

PR #573, opened 2022-06-14
Fix: use asset types to avoid breaking sorting

agoose77:
Currently, we break the priority sorting of assets by writing strings back to the asset lists. This PR ensures that the asset instance types (Stylesheet,  JavaScript) are used with a default priority to avoid breaking things.

I don't love this implementation, but it's probably good enough™

### DIFF
--- a/src/sphinx_book_theme/__init__.py
+++ b/src/sphinx_book_theme/__init__.py
@@ -104,16 +104,11 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
             ix = asset_context.index(asset)
         except ValueError:
             continue
-
-        asset_hash = _gen_hash(asset_source_path)
+        asset_obj = asset_context[ix]
 
         # Regenerate the asset type with the hashed filename
-        asset_obj = asset_context[ix]
-        asset_context[ix] = type(asset_obj)(
-            filename=f"{asset_sphinx_link}?digest={asset_hash}",
-            priority=asset_obj.priority,
-            **asset_obj.attributes,
-        )
+        asset_hash = _gen_hash(asset_source_path)
+        asset_obj.filename = f"{asset_sphinx_link}?digest={asset_hash}"
 
 
 def hash_html_assets(app, pagename, templatename, context, doctree):
--- a/src/sphinx_book_theme/__init__.py
+++ b/src/sphinx_book_theme/__init__.py
@@ -107,6 +107,7 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
 
         asset_hash = _gen_hash(asset_source_path)
 
+        # Regenerate the asset type with the hashed filename
         asset_obj = asset_context[ix]
         asset_context[ix] = type(asset_obj)(
             filename=f"{asset_sphinx_link}?digest={asset_hash}",
--- a/src/sphinx_book_theme/__init__.py
+++ b/src/sphinx_book_theme/__init__.py
@@ -90,28 +90,32 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
                 f"Asset {asset_source_path} does not exist, not linking."
             )
             continue
-            
-        asset_hash = _gen_hash(asset_source_path)
-        new_asset_sphinx_link = f"{asset_sphinx_link}?digest={asset_hash}"
         
         # CSS assets are stored in css_files, JS assets in script_files
         if asset.endswith(".css"):
-            new_asset = Stylesheet(new_asset_sphinx_link)
             asset_context = context.get("css_files", [])
         elif asset.endswith(".js"):
-            new_asset = JavaScript(new_asset_sphinx_link)
             asset_context = context.get("js_files", [])
         else:
             SPHINX_LOGGER.warn(
                 f"Unrecognised asset type, not hashing."
             )
+            continue
             
         # Find this asset in context, and update it to include the digest
         try:
             ix = asset_context.index(asset)
         except ValueError:
             continue
-        asset_context[ix] = new_asset
+
+        asset_hash = _gen_hash(asset_source_path)
+
+        asset_obj = asset_context[ix]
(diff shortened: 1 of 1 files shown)


src/sphinx_book_theme/__init__.py
```diff
@@ -68,6 +69,9 @@ def _gen_hash(path: str) -> str:
     return hashlib.sha1(path.read_bytes()).hexdigest()
 
 
+from sphinx.builders.html import JavaScript, Stylesheet
+
+
```
choldgraf, line 74:
```suggestion
```

src/sphinx_book_theme/__init__.py
```diff
@@ -81,20 +85,37 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
...
-            if asset_sphinx_link in context[asset_type]:
-                hash = _gen_hash(asset_source_path)
-                ix = context[asset_type].index(asset_sphinx_link)
-                context[asset_type][ix] = asset_sphinx_link + "?digest=" + hash
+        if asset.endswith(".css"):
+            asset_context = context.get("css_files", [])
+        elif asset.endswith(".js"):
+            asset_context = context.get("js_files", [])
+        else:
+            SPHINX_LOGGER.warn(f"Unrecognised asset type, not hashing.")
```
choldgraf, line 93:
```suggestion
            SPHINX_LOGGER.warn(f"Unrecognised asset type: {asset}. Not hashing.")
```

src/sphinx_book_theme/__init__.py
```diff
@@ -81,20 +85,37 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
...
+            SPHINX_LOGGER.warn(
+                f"Asset {asset_source_path} does not exist, not linking."
+            )
+            continue
+
+        # Find this asset in context, and update it to include the digest
+        try:
+            ix = asset_context.index(asset)
+        except ValueError:
+            continue
```
choldgraf, line 109: Why not raise an error here? I feel like it would be pretty confusing if it just skipped it because of some bug in the name matching or something

src/sphinx_book_theme/__init__.py
```diff
@@ -81,20 +85,37 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
...
-            if not asset_source_path.exists():
-                SPHINX_LOGGER.warn(
-                    f"Asset {asset_source_path} does not exist, not linking."
-                )
-            # Find this asset in context, and update it to include the digest
-            if asset_sphinx_link in context[asset_type]:
-                hash = _gen_hash(asset_source_path)
-                ix = context[asset_type].index(asset_sphinx_link)
-                context[asset_type][ix] = asset_sphinx_link + "?digest=" + hash
+        if asset.endswith(".css"):
```
choldgraf, line 88:
```suggestion
        # Use the html context list of assets so we replace it in-place and retain ordering
        if asset.endswith(".css"):
```

src/sphinx_book_theme/__init__.py
```diff
@@ -7,6 +7,7 @@
 from docutils.parsers.rst.directives.body import Sidebar
 from docutils import nodes as docutil_nodes
 from sphinx.application import Sphinx
+from sphinx.builders.html import JavaScript, Stylesheet
```
choldgraf, line 10:
```suggestion
```

it seems these aren't used?

src/sphinx_book_theme/__init__.py
```diff
@@ -81,20 +85,37 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
...
+
+        # Find this asset in context, and update it to include the digest
+        try:
+            ix = asset_context.index(asset)
+        except ValueError:
+            continue
+
+        asset_hash = _gen_hash(asset_source_path)
+
+        asset_obj = asset_context[ix]
```
choldgraf, line 113:
```suggestion
        # Regenerate the asset type with the hashed filename
        asset_obj = asset_context[ix]
```

src/sphinx_book_theme/__init__.py
```diff
@@ -81,20 +85,37 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
...
+        asset_context[ix] = type(asset_obj)(
```
choldgraf, line 114: Is it not possible to update the filename in-place for these assets? We have to re-create it?

choldgraf (commented):
Ahh this one has been bugging me forever. Good discovery and thanks for the patch!

I left a few comments in there, but it LGTM in general. Let me know what you think.

src/sphinx_book_theme/__init__.py
agoose77: I'm preserving the existing behaviour. Would you rather that I raise here?

choldgraf: Let's try raising an error and seeing if this happens in an annoying way? Though if this spirals into a pain in the butt to work around, I'm fine leaving that to an issue to track w/o needing it in thi sPR

choldgraf:
Also a quick reminder to run the pre-commit install before your next commit so that you can catch pre-commit errors that way!

src/sphinx_book_theme/__init__.py
agoose77: I'll do the same for the other warning, then

agoose77:
> Also a quick reminder to run the pre-commit install before your next commit so that you can catch pre-commit errors that way!

Hehe, I wrote this PR at midnight and avoided most best practices ;)

choldgraf (approved):
This looks good to me as long as tests are happy!

agoose77:
@choldgraf to be honest, I'm running out of cycles to work on things like this. It looks like this code is never tested, so coverage is complaining. Is this something you have any time to look at?

choldgraf:
Don't sweat it, I can shepherd it in if you don't have the cycles 👍

PR #579, opened 2022-06-26
Stylize Read the Docs as per guidelines

pradyunsg:
https://read-the-docs-guidelines.readthedocs-hosted.com/branding.html#brand-name


pradyunsg:
(Yay on `sphinx-theme-builder`, which means I don't need to regenerate stuff locally!)

choldgraf (approved):
Thanks for catching this 🙏

humitos:
He! I arrived here to change the same 😄 

It seems that https://jupyterbook.org/en/latest/intro.html is still using the old version of the theme.

PR #582, opened 2022-07-07
Add basthon launch button

fortierq:
This PR adds a launch button for Basthon.

[Basthon](https://notebook.basthon.fr) is a nice project to execute a notebook client-side (inside the browser), using Pyodide. Similar to Jupyterlite, but I found it more stable for now and usable with several languages (SQL, OCaml...).


fortierq:
Closed for #595

PR #585, opened 2022-07-12
ENH: Improve Chinese (Taiwan) and Chinese (China) translation

whyjz:
I edited the translations to make them look more natural and better match the context.

PS. I am a native Chinese (Taiwan) speaker and have a decent understanding of Chinese (China).


choldgraf (approved):
This is so helpful, thanks very much @whyjz - it is really hard to keep translations up-to-date and accurate 🙏

PR #586, opened 2022-07-13
Remove last-child zero-margin on <dd>'s

benjaoming:
Hi there :wave: 

The style for dl's seems to originate from https://github.com/executablebooks/sphinx-book-theme/pull/191 and following some refactors, I don't see any changes or reasons why this margin seems so off.

I suppose it's unintentional that a definition is closer to the succeeding term than the term its actually a definition for?

Before:

image

After:

image


choldgraf:
This looks great to me - better UI and 6 removed lines! Thanks for the improvement

Issue #575, opened 2022-06-18
Update pydata theme to 0.10


pradyunsg:
### Context

The PyData Sphinx Theme has made a bunch of improvements and modifications in their latest release, and it has a release candidate out for v0.10 (two minor version bumps from our current pin). Here's the rough changelog:

- https://github.com/pydata/pydata-sphinx-theme/releases/tag/v0.10.0rc1

### Things to update / change

Here are some of the big things that have changed, and which we should figure out how to incorporate here:

- We upstreamed a bunch of CSS for extensions, fixing bugs, etc. I suspect that we can reduce a lot of the custom CSS in this theme now.
- We standardized several sections using the `basic-ng` naming. E.g. `bd-sidebar` became `bd-sidebar-primary`. We might need to re-work some of our CSS rules accordingly. I think we might also want to just re-use the same class prefix for this theme (`bd-`) so that we can inherit some of the same CSS behavior.
- New theme sections were added. In particular there's now an `announcements.html` so we might be able to use that instead of re-inventing our own here.
- There is now a Dark Mode in the pydata theme and we need to find ways to inherit it
- There is new search bar functionality via a pop-up instead of a field. We could either make our current field "behave" like a pop-up, which would make it easier to extend to other search services. Or we could just re-use the button
- We need a place for the theme toggle. I think the easiest thing would be to support the `icon-links` component in our sidebar. People could also put site-wide links there if they wished (ref #285)

### Proposal

We can use this PR to make the version bumps. I started it, but **anybody should feel free to take it over and/or push to it**. I think @AakashGfude might be interested in working on this?

- #574

edit: Issue hijacked by Chris to have more detail in it

SilverRainZ:
wow! Looking forward to a new release.

PR #574, opened 2022-06-16
[WIP] Update pydata theme

choldgraf:

PR #597, opened 2022-07-28
Update pydata theme to 0.10

AakashGfude:
This PR updates the theme to synchronize with pydata-sphinx-theme~=0.10. The aim here is to just refactor the existing code to harmonize with the latest pydata. So, no new features will be added. Although a lot of CSS rework would be done here, apart from HTML and CSS. 

- [x] CSS rework to remove redundancy and mimic the existing sphinx-book-theme style, and use variables wherever possible.
- [x] Handling footer-content in pydata-sphinx-theme.  Latest pydata although following most of the skeleton of sphinx-basic-ng does not seem to have a section for footer-content. Which is needed in sphinx-book-theme : 
   
**footer-content in sbt**:
Screen Shot 2022-07-28 at 11 12 22 am

  **for reference in s-b-ng**:

Screen Shot 2022-07-28 at 11 13 52 am

We can probably just use a hacky one in this PR and update it in relevant sources in upcoming cycles. 

EDIT: Done in https://github.com/pydata/pydata-sphinx-theme/pull/861

- [x] Javascript not broken with updated HTML.
- [x] restructuring HTML to inherit as much as possible. 

fixes https://github.com/executablebooks/sphinx-book-theme/issues/575


choldgraf:
Just a note that some of this has already been done in the PR below, but if you think it's better to start a fresh PR I am +1. Will link below if it's useful

- https://github.com/executablebooks/sphinx-book-theme/pull/574

I think we should use this as an opportunity to simplify and reduce as much as we can...there were many special-cases in CSS, HTML structure, etc that we had in this theme, that we have since upstreamed to the pydata theme, so we should be able to slim things down here a lot more.

AakashGfude:
> Just a note that some of this has already been done in the PR below, but if you think it's better to start a fresh PR I am +1. Will link below if it's useful
> 
>     * [[WIP] Update pydata theme #574](https://github.com/executablebooks/sphinx-book-theme/pull/574)
> 
> 
> I think we should use this as an opportunity to simplify and reduce as much as we can...there were many special-cases in CSS, HTML structure, etc that we had in this theme, that we have since upstreamed to the pydata theme, so we should be able to slim things down here a lot more.

Ahh yes, Thank you for linking the PR.

choldgraf:
@AakashGfude let's try to add in the content footer in the PyData theme as an empty template for now. That way sub themes can override it even if it's not used in the PyData theme.

pyproject.toml
```diff
@@ -22,8 +22,8 @@ readme = "README.md"
 
 requires-python = ">=3.7"
 dependencies = [
-  "sphinx>=3,<5",
-  "pydata-sphinx-theme~=0.8.0",
+  "sphinx>=3,<6",
+  "pydata-sphinx-theme~=0.10.0rc1",
```
choldgraf, line 26: I'd try rc2 - that makes some more changes

AakashGfude:
@choldgraf I think this is good to go. Sorry, this PR became huge, but it just has the necessary changes to adopt to pydata ~=0.10.1

src/sphinx_book_theme/assets/styles/abstracts/_variables.scss
```diff
@@ -50,9 +52,15 @@ $border-thin: 1px solid rgba(0, 0, 0, 0.1);
   --pst-color-admonition-important: var(--pst-color-info);
 
   --pst-color-warning: 245, 162, 82;
+  --pst-color-admonition-caution: var(--pst-color-warning);
+  --pst-color-admonition-attention: var(--pst-color-warning);
+
 
   --pst-color-danger: 230, 101, 129;
```
choldgraf, line 56: Didn't the `pst` variables get re-worked to be full `rgb(` values now instead of just numbers?

src/sphinx_book_theme/assets/styles/content/_admonitions.scss
```diff
@@ -1,20 +1,133 @@
-.admonition {
+.admonition, div.admonition {
   // Soften admonition styling from the PyData Theme
   border-radius: $box-border-radius;
   box-shadow: 0 0.2rem 0.5rem rgba(0, 0, 0, 0.05),
     0 0 0.0625rem rgba(0, 0, 0, 0.1);
-
+  padding: 0 .6rem .8rem;
+  border-color: rgba(var(--pst-color-admonition-default),1);
```
choldgraf: Why do we need to be setting the border colors etc here? Couldn't we just inherit from pydata?

src/sphinx_book_theme/assets/styles/content/_admonitions.scss
```diff
@@ -1,20 +1,133 @@
...
   // Admonition seealso styling
   &.seealso {
     border-color: #28a745;
-
-    .admonition-title {
+    > .admonition-title {
       background-color: rgba(40, 167, 70, 0.1);
 
-      &:before {
+      &:after {
```
choldgraf, line 109: same for here, couldn't we just inherit?

src/sphinx_book_theme/assets/styles/content/_margin.scss
```diff
@@ -5,14 +5,14 @@
 // So the width we want is:
 // max-width * x = 100% - max-width, so we solve for x.
 // Will be a decimal
-$content-margin-gutter: 3%;
-$content-margin-width: percentage(
-    (100% - $content-max-width) / $content-max-width
+$content-margin-gutter: 2rem;
+$content-margin-width: (
+    (88rem - $content-max-width)
```
choldgraf: we should re-use a variable here if `88rem` is used elsewhere, or document what `88rem` comes from

src/sphinx_book_theme/assets/styles/sections/_article.scss
```diff
@@ -9,7 +9,8 @@
       position: relative; // needed for postion:sticky header-article
       overflow-x: unset; // needed for postion:sticky header-article
       padding: 0%;
-      max-width: 50em;
+      max-width: $content-max-width;
```
choldgraf: what does it look like if we just inherit what the pydata theme does rather than defining our own?

choldgraf (commented):
This looks like a good start, thanks for all of this conversion! A few general thoughts and things I noticed:

- I was surprised that there was a net increase in lines of code. Our goal should be to re-use as much as possible from the PyData Theme and not re-code our own rules etc. Can you take a pass and see what you can remove from this theme such that it still looks good?
- the right table of contents icon is now showing up on widescreen: image
- search the docs is now very faint on the sidebar: image
- There seems to be an intermediate screen width with no margins: image
- There's now a left margin when the primary sidebar is collapsed. Is that intentional? image

AakashGfude:
> This looks like a good start, thanks for all of this conversion! A few general thoughts and things I noticed:
> 
>     * I was surprised that there was a net increase in lines of code. Our goal should be to re-use as much as possible from the PyData Theme and not re-code our own rules etc. Can you take a pass and see what you can remove from this theme such that it still looks good?

Yes, the lines have increased, unfortunately. But mostly because I had to introduce a lot of CSS here, like in src/sphinx_book_theme/assets/styles/content/_admonitions.scss . Because it was no longer getting styled by the new `pydata-sphinx-theme`.

AakashGfude:
>     * the right table of contents icon is now showing up on widescreen: image

Thanks for pointing this out. Have corrected it now.

>     * search the docs is now very faint on the sidebar: image
>       
>       * We may want to just re-use the search bar button instead of having a dedicated search bar as well
> 
>     * There seems to be an intermediate screen width with no margins: image
> 
>     * There's now a left margin when the primary sidebar is collapsed. Is that intentional? image

AakashGfude:
>     * There seems to be an intermediate screen width with no margins: image

I have corrected it now. The style was inherited from pydata-sphinx-theme and pydata-sphinx has a breakpoint of 1200px, where some UI features change, apart from the usual 768px. But in sphinx-book-theme, there's only one breakpoint of 768px which directly shifts it to mobile view. 
This is also one of the reasons why the lines of code in the PR are huge, as I have overwritten some of the design decisions of pydata, as I did not want to introduce any new design changes in this PR for sphinx-book-theme. 

>     * There's now a left margin when the primary sidebar is collapsed. Is that intentional? image

Can't see it in the image you posted or maybe screen color is hiding it?

src/sphinx_book_theme/assets/styles/abstracts/_variables.scss
AakashGfude: That is true. But, I will have to rewrite this whole file for that then for consistency. As other variables are using numbers. You reckon I should do it in this PR?

src/sphinx_book_theme/assets/scripts/index.js
```diff
@@ -56,7 +56,7 @@ var toggleFullScreen = () => {
  * the screen.
  */
 var scrollToActive = () => {
-  var navbar = document.getElementById("site-navigation");
+  var navbar = document.getElementsByClassName("bd-sidebar-primary")[0];
```
choldgraf, line 59: I think that we should be able to delete the whole scrolltoactive section, since that code exists in the pydata theme:

https://github.com/pydata/pydata-sphinx-theme/blob/cf8a947c695211600d928bbb39bcd74b6f11ce8a/src/pydata_sphinx_theme/assets/scripts/index.js#L155

src/sphinx_book_theme/assets/scripts/index.js
```diff
@@ -230,3 +240,4 @@ sbRunWhenDOMLoaded(initTooltips);
 sbRunWhenDOMLoaded(scrollToActive);
 sbRunWhenDOMLoaded(initTocHide);
 sbRunWhenDOMLoaded(initRTDObserver);
+sbRunWhenDOMLoaded(addNoPrint);
```
choldgraf, line 243: will this work if we print via headless chromium? That's what we use for HTML-to-PDF in jupyter book

src/sphinx_book_theme/assets/styles/content/_admonitions.scss
```diff
@@ -1,20 +1,141 @@
...
         color: #28a745;
         content: "\f064";
       }
     }
   }
+  // Admonition warning styling
+  &.warning {
```
choldgraf, line 116: What happens if we just delete our `_admonitions.scss` file entirely? IMO that should be our goal.

src/sphinx_book_theme/assets/styles/pages/_search.scss
```diff
@@ -3,45 +3,50 @@
  */
 
 // Turn off the extra search UI at the top of the page
-#search-documentation {
-  display: none;
-  & ~ p,
-  ~ form {
+.bd-search-container {
```
choldgraf, line 6: Could we re-use the icon search pattern in the pydata theme instead of defining our own text field for it? For example, we could put the search button in the sidebar along with any icons that were site-wide and not specific to the page:

image

src/sphinx_book_theme/assets/styles/abstracts/_variables.scss
choldgraf: I think so, otherwise we are going to have two different CSS variable structures, one that uses `rgb(` and another that just defines numbers. I think we should just standardize on using `rgb(`

choldgraf (commented):
Thanks @AakashGfude for making more progress -  here are another few thoughts and comments after taking another pass.

> Yes, the lines have increased, unfortunately. But mostly because I had to introduce a lot of CSS here

I think that our goal in this PR, and in the theme in general, is "reduce the number of lines of code as much as possible, and re-used as much as possible from the PyData theme out of the box". In your next iterations on this PR can you try and follow that principle in removing things that we special-case here?

An example, I think basically any block-level content styling we can remove and just re-use whatever the pydata theme does. The kinds of things we'll need to style ourselves are the things that are really unique to this theme, like the dropdown table of contents

> This is also one of the reasons why the lines of code in the PR are huge, as I have overwritten some of the design decisions of pydata, as I did not want to introduce any new design changes in this PR for sphinx-book-theme.

I think for most design changes, it is probably fine unless we need to make small tweaks. We should inherit as much as possible from the pydata theme (e.g. admonition style, code block style, other block element styles, etc). We don't have the maintenance capacity to maintain a bunch of custom styles that are just for the book theme IMO, so the point of using pydata is to not have to re-invent the wheel.

> Can't see it in the image you posted or maybe my screen color is hiding it?

Hmm I can't see if anymore, maybe that one got fixed in a commit.

I'm happy to chat about this in a meeting if that's helpful. I think we should define a goal for this PR and then continue to iterate from there.

choldgraf:
ping @AakashGfude - when will you have another chance to look at this? I think we should get this upgrade done as quickly as we can because we're just going to keep drifting away from the pydata theme otherwise.

If there are questions or uncertainties you'd like to resolve via a meeting I'm happy to have a conversation if that's helpful.

AakashGfude:
> ping @AakashGfude - when will you have another chance to look at this? I think we should get this upgrade done as quickly as we can because we're just going to keep drifting away from the pydata theme otherwise.
> 
> If there are questions or uncertainties you'd like to resolve via a meeting I'm happy to have a conversation if that's helpful.

@choldgraf I am on this one now. Sorry was away, took a cheeky trip to bail. Will work on this, and see if all conversations can be resolved.

src/sphinx_book_theme/assets/styles/content/_admonitions.scss
AakashGfude: Hi @choldgraf, thank you for your input on this. Very valuable point of inheriting as much as we can without paying too much heed to backward compatibility styling. I have removed the `_admonitions.scss` and also some colors defined in 'src/sphinx_book_theme/assets/styles/abstracts/_variables.scss', which we will directly inherit from pst.

src/sphinx_book_theme/assets/scripts/index.js
AakashGfude: how is it supposed to work again? the primary sidebar should scroll automatically if the active page is off-screen right?

src/sphinx_book_theme/assets/styles/pages/_search.scss
AakashGfude: @choldgraf actually we are using the html defined in pydata only for this. Just adding our own styles for positioning it. https://github.com/pydata/pydata-sphinx-theme/blob/main/src/pydata_sphinx_theme/theme/pydata_sphinx_theme/components/search-field.html. But I am okay with using the pydata search pattern. Should we do it in a separate PR, as its more like a new enhancement? Or you want it here?

src/sphinx_book_theme/assets/scripts/index.js
AakashGfude: Hi @choldgraf , seems to be working. Primary sidebars and announcement banners etc. are hidden in pdfhtml build.

src/sphinx_book_theme/assets/scripts/index.js
AakashGfude: @choldgraf I checked using master, and this doesn't seem to be working https://github.com/executablebooks/sphinx-book-theme/pull/609 . Did it work for you?

src/sphinx_book_theme/assets/styles/content/_margin.scss
AakashGfude: have reverted this back

src/sphinx_book_theme/assets/styles/sections/_article.scss
AakashGfude: have reverted this back

AakashGfude:
@choldgraf let me know what you think on this. I have not done the https://github.com/executablebooks/sphinx-book-theme/pull/597#discussion_r969396971 here. The reason is in that thread. Apart from that, I think the rest of the queries have been resolved. 

Also this https://github.com/executablebooks/sphinx-book-theme/pull/597#discussion_r969384783. Not able to test it, or I am doing something wrong?

choldgraf:
This is looking better! A few quick comments below:

- Fine to do the icon stuff in a follow-up PR.
- The article header and the sidebar have a break in the shadow now. Can we avoid this break somehow? image

Anything else you want to tackle here? If not then I suggest we:

- Merge any small PRs that are open
- Make a release
- Merge this PR so that we have time to spot bugs etc

AakashGfude:
> * The article header and the sidebar have a break in the shadow now. Can we avoid this break somehow?
>   image

@choldgraf The main reason is actually because, earlier the secondary sidebar was inside header-article:

Screen Shot 2022-10-01 at 11 54 29 pm

Now we have moved it outside. sidebar-secondary and header-article are separate items. And box shadow is very fussy. So we can achieve it probably, but will be a bit messy. I will look at some hacks if any. 



> * I think our ReadTheDocs JS stopped working, probably a CSS selector change in the JS code:

What part is not working? I am able to open the menu and navigate to options, and use the search bar. In the preview.

AakashGfude:
> Anything else you want to tackle here? If not then I suggest we:

I think this PR looks sufficient from my side.

> * Merge any small PRs that are open
> * Make a release

Let me know, if I can help with this.

> * Merge this PR so that we have time to spot bugs etc

Sounds good.

pradyunsg:
FYI: 0.10 breaks the preview on sphinx-themes.org due to a bug: https://github.com/sphinx-themes/sphinx-themes.org/issues/105

choldgraf:
@AakashGfude - sorry for the slow turnaround, I have a bunch of other stuff going on. @mmcky any chance that you could give the docs a look as well and make sure nothing looks off?

A few things that I noticed:

### Dark theme can be auto-set by system

The pydata theme has a default to "use the system preference" for light/dark theme. That means that we might accidentally trigger the dark theme. For example, here's what it looks like when my computer has dark theme on by default:

image

I think there are two ways to approach this:

1. Set the default to light, [by following these instructions](https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/light-dark.html#configure-default-theme-mode)
2. Add the CSS + button for a dark theme in this PR

### Full width cells are off screen

The full-width cells seem to spill off screen:

image

The margin notes also seem to spill over and have a littele extra white space on the left:

image

choldgraf:
A few more visual bugs. Can you please go through [the kitchen sink](https://sphinx-book-theme--597.org.readthedocs.build/en/597/reference/kitchen-sink/index.html) and [the theme-specific elements page](https://sphinx-book-theme--597.org.readthedocs.build/en/597/reference/special-theme-elements.html) and confirm that everything looks normal. That is the easiest way to spot these bugs. I'd like to get this PR merged quickly and I know that we will need to keep iterating, but I at least want to avoid introducing an obvious visual bug into our docs.

- [ ] Admonitions still have a weird margin space: image
- [ ] Code blocks with a title have extra margin at the bottom: image

AakashGfude:
@choldgraf thanks for reporting this, and looking through all this patiently. I think the ones notified in the last comment have been resolved with the latest push. I will look through those pages for any visual anomalies.

choldgraf:
Sounds good @AakashGfude - let me know when you've taken a close look and think it is ready to merge, and then I can give a final look-over.

AakashGfude:
@choldgraf from my inspection of the pages, I think this PR looks good. 

Just pointing out some visual changes, which are inherited from **pydata-sphinx-theme**. The changes inherited looks better to me.

**Left**: original                                                                                                         **right**: this PR

1) space between caption and legend.
<p float="left">
Screen Shot 2022-10-23 at 5 46 50 pm
Screen Shot 2022-10-23 at 5 46 33 pm
</p>

2) The position of caption for list tables are above the list tables.
<p float="left">
Screen Shot 2022-10-24 at 8 45 24 pm
Screen Shot 2022-10-24 at 8 26 12 pm
</p>

3) double ticks are bordered. 
<p float="left">
Screen Shot 2022-10-24 at 8 51 19 pm
Screen Shot 2022-10-23 at 2 05 40 pm
</p>

stevepiercy:
Sorry to jump in late with feedback. I like the improvements and customizations made so far.

I found a few more discrepancies that are important to our documentation projects.

The search box hint text is very light and difficult for me to read.

## Original

search-original

## Proposed

search-new

The `seealso` admonition's icon and color changed to the default `info` admonition. I prefer the green curvy arrow icon.

## Original

seealso-orig

## Proposed

seealso-new

The contents menu no longer highlights the final heading of a page when scrolled to the end.

## Original

contents-menu-orig

## Proposed

contents-menu-new

The `hint` and `important` admonitions have changed from yellow to green. I think they should be reverted to yellow.

## Original

hint-important-orig

## Proposed

hint-important-new

The `tip` admonition has been changed from yellow to green. I think it should be reverted.

## Original

tip-orig

## Proposed

tip-new

choldgraf:
Thanks @stevepiercy ! A few quick responses from me on next steps for these suggestions:

- The search box hint text is very light and difficult for me to read.
  - Should be solveable with `placeholder` CSS: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#placeholder_2
- The seealso admonition's icon and color changed to the default info admonition. I prefer the green curvy arrow icon.
  - Let's fix this one upstream - I opened an issue to track it and am happy to review PRs to implement a fix:
  - https://github.com/pydata/pydata-sphinx-theme/issues/1036
- The contents menu no longer highlights the final heading of a page when scrolled to the end.
  - I suspect that this is because the end section in this case is relatively short - it is highlighting a section once it is above X% of the page, and I bet it's just not high enough to trigger that here. Maybe we can tweak the UI/UX a bit, but it doesn't feel to me like a "bug" per se.
- The tip admonition has been changed from yellow to green. I think it should be reverted.
  - This got changed in the upstream theme in order to reduce the number of unique color variables being used. If we want to use yellow we'll need to define our own color variable in both light and dark modes and apply it to these two admonition types. I feel like we could open an issue and track that one so that it doesn't block this PR.

AakashGfude:
Thank you for the inputs @stevepiercy, and for the replies @choldgraf. The latest push should have the search field text covered. I agree with the admonition styling being done upstream. 
The highlighting on the end of the page, i reckon as mentioned by @choldgraf is subjective it seems. In my laptop, it is working fine:
Screen Shot 2022-10-25 at 11 58 50 pm

stevepiercy:
Thanks for the responses. I agree on the course forward on all points. I will work on the new issues next week. Full speed ahead for this PR!

choldgraf (approved):
OK I think that this one is ready to go - I bet that we will still need to make some tweaks over time, and I also think it'd be great to try and take another pass at removing more customizations here to just re-use the pydata theme structure as much as possible, but this is a big complex PR and it'll be out of date quickly if we merge anything in. @AakashGfude what do you think?

AakashGfude:
@choldgraf I agree. Let's merge this one. And continue the tweaks over time.

stevepiercy:
On second thought, I think the `tip` and `hint` admonitions are OK as green, but I think blue would be better. The color green has a universal context of "go", whereas blue's is "information".

I think `important` should be yellow, not green, as it is similar in purpose and context to `caution`.

I can see that this theme inherits the colors from pydata: https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html

I could open an issue in the pydata theme to discuss, but I wanted to get @choldgraf's opinion first. Please let me know. Thank you!

choldgraf:
I think that makes sense to me - let's discuss in the pydata theme and see where that goes - what do you think?

stevepiercy:
Done! https://github.com/pydata/pydata-sphinx-theme/issues/1040

kloczek:
Any chance to tag new version soon? 🤔

SilverRainZ:
Kindly ping, any plan to make a new release? :D

choldgraf:
I think there are some more breaking changes we'll need to make here: 

- https://github.com/executablebooks/sphinx-book-theme/issues/630

My feeling is that we should get that done at least as an MVP before we release, so that we can simplify this one as much as possible and start to make it much easier to update in the future

Issue #608, opened 2022-09-04
Show active item of main navigation


ksuess:
### Describe the bug

The active navigation item is not visible.

### Reproduce the bug

Given a navigation so long that it does not fit into viewport height.
Scroll down and select one of the last menu items.
This active menu item is out of view after the new page is loaded.

### List your environment

current sphinx-book-theme 0.3.3

PR #609, opened 2022-09-04
FIX: Scroll to active navigation item

ksuess:
Fix https://github.com/executablebooks/sphinx-book-theme/issues/608

### DIFF
--- a/src/sphinx_book_theme/assets/scripts/index.js
+++ b/src/sphinx_book_theme/assets/scripts/index.js
@@ -57,6 +57,7 @@ var toggleFullScreen = () => {
  */
 var scrollToActive = () => {
   var navbar = document.getElementById("site-navigation");
+  var navbar_scrollable = navbar.children[0];
   var active_pages = navbar.querySelectorAll(".active");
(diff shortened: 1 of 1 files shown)


ksuess:
It would be great to have this fix merged and released before the big upgrade to current pydata-theme. @choldgraf, could you have a look?

choldgraf (approved):
This looks good to me - I was thinking that we should wait and see if the pydata theme upgrade fixes this, but I agree that'll be a bit of a bigger change and we might as well push out this bugfix in the meantime 👍 

- ref: https://github.com/executablebooks/sphinx-book-theme/pull/597

ksuess:
Thank you @choldgraf for reviewing and merging. 
The upgrade to pydata-sphinx-theme~=0.10 will bring breaking changes. Maybe a 0.3.x release before this upgrade would be a good idea. What do you think?

choldgraf:
agreed - I created an issue to cut a patch release here:

- https://github.com/executablebooks/sphinx-book-theme/issues/611

However I don't have much time for this so I welcome others to give a shot at releasing

PR #610, opened 2022-09-13
DOC: Update kitchen sink

choldgraf:
This updates our kitchen sink to the latest version that's on the sphinx themes gallery. It does this by copying over the script we use in the pydata theme so they're basically the same.


choldgraf:
Merging this in since it brings us in-line with the pydata theme's sink and will make it easier to compare the differences with #575

PR #613, opened 2022-09-19
Relax requirement that launch button URLs must point to GitHub repositories

tschm:
Generalize the _split_repo_url. Still works as exactly as before for GitHub urls but doesn't fall over for GitLab.


choldgraf (commented):
This seems reasonable, though I'm hesitant to just accept any arbitrary URL in case people don't understand that only a Binder-supported URL will work. Could we instead check for `github.com/` or `gitlab.com/` is in the URL and if not then provide a helpful error message?

tschm:
I was hoping I could point to our internal Jupyter-Hub server. There is no chance I can rely on Binder.

choldgraf:
Gotcha - can you confirm that this actually works for GitLab or other repositories?

PR #616, opened 2022-09-28
Add skip to main content link

andesol:
Add a simple jump to main content link on top of the page (only visible when you first press tab) so users can skip the left navigation

image

### DIFF
--- a/src/sphinx_book_theme/assets/styles/components/_buttons.scss
+++ b/src/sphinx_book_theme/assets/styles/components/_buttons.scss
@@ -117,3 +117,18 @@ div.header-article-main {
     display: none;
   }
 }
+
+.skip-link {
+  display: block;
+  left: -999px;
+  position: absolute;
+  top: -999px;
+
+  &:focus {
+    left: 0;
+    padding: 0.2rem 0.3rem;
+    position: relative;
+    top: 0;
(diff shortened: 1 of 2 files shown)


choldgraf:
Could you provide some more information about what you mean by "skip the left navigation"? That should all be hidden by default. Is this for screen readers and accessibility purposes?

chrisjsewell:
It's this @choldgraf https://youtu.be/VUR0I5mqq7I

choldgraf:
So this one has actually now been implemented in the upstream theme (pydata-sphinx-theme):

- https://github.com/pydata/pydata-sphinx-theme/issues/966

I think that we can close this PR since we'll inherit this feature from the upstream, though we should double-check that it is in-fact inherited

many thanks @andesol for surfacing this issue and providing an implementation - I think that the person in the pydata theme used it for inspiration there

choldgraf:
I'm gonna close this one as we now have an implementation we'll inherit from the pydata theme!

Here's the PR where this was merged upstream:

- https://github.com/pydata/pydata-sphinx-theme/issues/966

And here's the PR that should bring this in, assuming that a PyData theme release is made before it is merged:

- #640

jonnew:
Is there any way to disable this in conf.py?

PR #620, opened 2022-10-07
Bump the version of pytest for compatibility with python 3.10

ZedThree:
Pytest 6.2.5 is the first version that supports python 3.10: https://pytest.org/en/latest/changelog.html#pytest-6-2-5-2021-08-29

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,7 @@ test = [
     "beautifulsoup4>=4.6.1,<5",
     "coverage",
     "myst-nb~=0.13.2",
-    "pytest~=6.0.1",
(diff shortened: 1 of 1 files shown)


AakashGfude:
Hi @ZedThree, I would suggest bump the version of pytest more, and pytest-cov as well. Like here: https://github.com/executablebooks/MyST-NB/blob/master/pyproject.toml#L103

ZedThree:
Should I bump `myst-nb` version at the same time?

AakashGfude:
@ZedThree, we will do that in a separate PR. It needs some additional testing.

PR #626, opened 2022-10-23
Open launch buttons in new window

guillaume-haerinck:
## What  does this PR do?

It adds the boolean option `open_launch_buttons_link_in_new_window` in the theme. When enabled, the buttons in the screenshot below have the `target="_blank"` option which opens the link in a new window instead of replacing the page.

image

## How was this PR tested?

I built the documentation with and without the option enabled and ran pre-commit conditions checks

## Why do you need it?

We embed our books inside some html iframe in order to easily switch between books on the [same website](https://learn-computer-graphics.com/mathematics/). It works great except for some links which will refuse to open inside an iframe.

I tried to change it with some Javascript but it's no use because of security restriction, thus I propose this new theme option.


choldgraf:
Thanks very much for this improvement! I think that the implementation looks good, however I have one question about whether we can simplify this a bit:

I was thinking about this UX, and want to propose that instead of making this a theme option and increasing our config surface area, we instead just change the _default behavior_ of launch buttons, so that they _always_ launch in a new window.

I think this is a reasonable change to make, because launch buttons are inherently taking you outside of the book itself into a new page or application. I find that many sites use this kind of "exit" action to spawn new tabs/windows.

What do you thinK?

guillaume-haerinck:
Hey ! Yes totally, I just updated the files :)

choldgraf (dismissed):
I think this makes sense - I'll hold off on merging for a day or two in case anybody else thinks that we _shouldn't_ do this, but LGTM. Thanks catching this!

guillaume-haerinck:
Hi @choldgraf :) Could you re-run the CI scripts ? Some of them were cancelled while others went through

choldgraf (approved):
This looks great to me, and tests are happy - thanks for noticing this and making the improvement!

PR #631, opened 2022-10-30
Add project URLs to pyproject.toml

rkdarst:
- Were missing (but in the readme), adding should be self-explanatory.
- These would have saved me a few seconds, so I may as well add them.
- Review: I'm not entirely sure if these are the common names in
  executablebooks.


choldgraf (approved):
This looks great - thanks for noticing and making the improvement!

PR #644, opened 2022-11-15
ENH: sphinx >=4,<6

AakashGfude:


choldgraf (approved):
Looks good to me - I did a quick look at the docs and they seemed to look fine to me. 

I noticed that the code cells in the right margin have an extra padding added to the left, but that may not be related to this change

image

AakashGfude:
> I noticed that the code cells in the right margin have an extra padding added to the left, but that may not be related to this change

 looks like the style is getting inherited from pydata sphinx theme, which we did not overwrite:

```
aside.sidebar>*
```
so not related to this PR. But the rest look okay to me.

mmcky:
Coordinating with https://github.com/executablebooks/sphinx-thebe/pull/57

PR #645, opened 2022-11-15
MAINT: Factor publish to pypi workflow into dedicated file

choldgraf:
Currently we have a workflow embedded in `tests.yml` that is called only if a new release is made. However this shows us as "not run" in our test suite in github and messes with the passing tests condition.

This PR factors that out into a dedicated workflow file that is run on `publish` and _calls_ our test suite before running the publish command. That should mean the "Publish to PyPI" action never gets run unless we are publishing.

cc @AakashGfude this should make it easier to quickly see if the tests are happy in PRs

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,33 @@
+# This will run every time a tag is created and pushed to the repository.
+# It calls our tests workflow via a `workflow_call`, and if tests pass
+# then it triggers our upload to PyPI for a new release.
+name: Publish to PyPI
+on:
+  release:
+    types: ["published"]
+
+jobs:
+  tests:
+    uses: ./.github/workflows/tests.yml
+  publish:
+    name: publish
+    needs: [tests] # require tests to pass before deploy runs
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
...
+on:
+  release:
+    types: ["published"]
+
+jobs:
+  tests:
+    uses: ./.github/workflows/tests.yml
+  publish:
+    name: publish
+    needs: [tests] # require tests to pass before deploy runs
```
AakashGfude, line 14: Do we also need 'docs-audit' here, or 'tests' is good enough you think?

AakashGfude (approved):
Thanks @choldgraf for this feature. Looks good to me. Apart from one comment above.

.github/workflows/publish.yml
choldgraf: This will run _all_ of the jobs that are in that workflow file, so it'll run `docs-audit` as well 👍

choldgraf:
thanks for the approve @AakashGfude - I think those two unreported tests are bugs in github not due to these changes, so lemme merge and see if that resolves it

PR #646, opened 2022-11-18
Enable JupyterHub launch button usage with GitLab

MP-bekh:
So far only GitHub repositories are supported to work with the JupyterHub launch button. This PR adds very little code that enables support for other servers as well such as GitLab as long as the repository name remains at the end of the repo url, i.e. https://examplegitlab.com/mygitlab/myrepo


MP-bekh:
Closed since this feature is already implemented in the newest version.

PR #658, opened 2022-12-12
MAINT: Move dependabot updates to monthly

choldgraf:
Dependabot is creating a lot of noise for small patch changes in our PRs. This changes it to monthly instead of weekly.

I don't think this one is controversial so will self-merge in a few days if I don't hear any objections.


choldgraf:
Merging since this is a tiny one

PR #660, opened 2022-12-14
MAINT: Update pre-commit versions and fix minor bugs in tests

choldgraf:
On systems that don't have the right version of GLIBC installed pre-commit will not work because of a missing dependency as described in this SO post:

- https://stackoverflow.com/questions/71939099/bitbucket-pipeline-error-installing-pre-commit-ts-lint/71940852#71940852

This fixes the version of node that is used so that this issue doesn't pop up, as per the guidelines in that issue.

It also avoids `ipython==8.7.0` in our tests to stop having a lexer bug that it introduced. 

I'll merge this one quickly so that we can fix pre-commit and get our tests passing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,10 +5,16 @@ exclude: >
     (?x)^(
       \.vscode/settings\.json|
     )$
+
+# Fix the node version to avoid a GLIBC error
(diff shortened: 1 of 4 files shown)


choldgraf:
Going to self-merge this one because it's just a minor fix to get our tests passing

PR #661, opened 2022-12-21
[FIX] Use logo url as is to allow for web urls.

feanil:
`logo_url` can sometimes be a fully qualified URL starting in Sphinx 4.
Adding the `_static` prefix breaks logos that are URLs.

### DIFF
--- a/src/sphinx_book_theme/theme/sphinx_book_theme/components/sidebar-logo.html
+++ b/src/sphinx_book_theme/theme/sphinx_book_theme/components/sidebar-logo.html
@@ -5,7 +5,7 @@
         {% set logo_url=logo %}
       {% endif %}
       {% if logo_url %}
-      <img src="{{ pathto('_static/' + logo_url, 1) }}" class="logo" alt="logo">
+      <img src="{{ logo_url|e }}" class="logo" alt="logo">
       {% endif %}
       {% if docstitle and not theme_logo_only %}
       <h1 class="site-logo" id="site-title">{{ docstitle }}</h1>


feanil:
@choldgraf the HTML builder handles this in Sphinx 4+

master: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/builders/html/__init__.py#L1254-L1255
v4.0.0: https://github.com/sphinx-doc/sphinx/blob/v4.0.0/sphinx/builders/html/__init__.py#L1191-L1195

src/sphinx_book_theme/theme/sphinx_book_theme/components/sidebar-logo.html
```diff
@@ -5,7 +5,7 @@
         {% set logo_url=logo %}
       {% endif %}
       {% if logo_url %}
-      <img src="{{ pathto('_static/' + logo_url, 1) }}" class="logo" alt="logo">
+      <img src="{{ logo_url|e }}" class="logo" alt="logo">
```
pradyunsg, line 8: Why is this being escaped? Isn't it an already-encoded URL?

pradyunsg:
x-ref https://github.com/pradyunsg/furo/pull/253 where I'd discovered this as well. :)

src/sphinx_book_theme/theme/sphinx_book_theme/components/sidebar-logo.html
feanil: @pradyunsg I don't think that's guaranteed since the `logo_url` value comes from the user in conf.py and the code I linked to above doesn't seem to be escaping anything.

pradyunsg: You're correct and I have to go fix something in Furo. :)

x-ref: https://github.com/sphinx-doc/sphinx/blob/6e6b9ad45194177811551c77f5f4040213d1c661/sphinx/themes/basic/layout.html#L56

choldgraf (approved):
Many thanks for this fix and for linking to the sphinx code where this is now handled! :-)

feanil:
Thanks for the quick reviews @choldgraf  and @pradyunsg,  when do you guys think this will be released in a new version?

choldgraf:
There's a release candidate out now and this will make it into the next RC i think

melund:
@fenil/@choldgraf The master version of sphinx-book-theme doesn't use the [sidebar-logo.htm](https://github.com/executablebooks/sphinx-book-theme/blob/master/src/sphinx_book_theme/theme/sphinx_book_theme/components/sidebar-logo.html)  anymore, but instead the [navbar-logo.html](https://github.com/pydata/pydata-sphinx-theme/blob/main/src/pydata_sphinx_theme/theme/pydata_sphinx_theme/components/navbar-logo.html) from the pydata theme. 

That version does check that the if the string is an URL, but if not it prepends `_static/` which causes me to see it prepended twice. I could maybe specify  
image

a workaround is to specify the logo using the PyData logo config. I guess that prevents sphinx from also prepending the `_static`: 

```
html_theme_options: Dict[str, Any] = {
      ...,
    "logo": {
      "image_light": "logo.svg",
      "image_dark": "logo.svg",
   },
```

feanil:
@melund it sounds like maybe this same fix needs to be made in [navbar-logo.html](https://github.com/pydata/pydata-sphinx-theme/blob/main/src/pydata_sphinx_theme/theme/pydata_sphinx_theme/components/navbar-logo.html) as well?

Issue #667, opened 2023-01-05
Anchor links have incorrect offset from top of screen


choldgraf:
When you click an in-page link (e.g. an anchor link) the link target is partially blocked by the article header.

This is because we are over-riding the `header-height` here to be 0px:

https://github.com/executablebooks/sphinx-book-theme/blob/8e087097e90c530440a25f724423faf6c99c2c37/src/sphinx_book_theme/assets/styles/abstracts/_variables.scss#L39

The pydata theme is setting `scroll-padding-top` to be relative to this height, which is why setting the height to 0 is causing this to overlap.

mathbunnyru:
A few other things I noted during comparison:

pre-release: https://jupyter-docker-stacks--1855.org.readthedocs.build/en/1855/using/recipes.html
stable: https://jupyter-docker-stacks.readthedocs.io/en/latest/using/recipes.html

1. In the pre-release version anchers are grey (they are not black). I liked black version better, but it might be my taste.

<img width="351" alt="Screenshot 2023-01-05 at 19 08 23" src="https://user-images.githubusercontent.com/12270691/210813182-0d51e0bd-bfae-495e-985d-ecb225b03927.png">
<img width="341" alt="Screenshot 2023-01-05 at 19 08 27" src="https://user-images.githubusercontent.com/12270691/210813189-a287f75c-74c5-4f0e-99bb-dd42b5265720.png">

2. I liked the text and the centred logo in the left-up corner. Now there is only a logo and it's not centred.

<img width="273" alt="Screenshot 2023-01-05 at 19 12 24" src="https://user-images.githubusercontent.com/12270691/210813355-7f07dd7e-290f-4b53-9150-e2b7b9f12dd1.png">
<img width="262" alt="Screenshot 2023-01-05 at 19 12 28" src="https://user-images.githubusercontent.com/12270691/210813361-1c3c8d73-3fe7-4ddd-b049-b9cb67f4e81e.png">

3. "Search the docs..." and "⌘ + K" are clashing.

4. Italic text in the anchors adds an extra space in the `contents` section. I think it might have happened before, but it's visible now.

<img width="233" alt="Screenshot 2023-01-05 at 19 09 33" src="https://user-images.githubusercontent.com/12270691/210813670-0041af48-db45-4afb-80b9-c3a673102173.png">
[...]

PR #669, opened 2023-01-05
FIX: Scroll padding on top for anchor links

choldgraf:
- closes https://github.com/executablebooks/sphinx-book-theme/issues/667

The failing test is a lighthouse performance check, which is notoriously flaky and I think can be ignored

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,6 +27,9 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: [3.7, 3.8, 3.9]
+        # Only test the latest major release of Sphinx because otherwise we need to
+        # keep multiple versions of regression tests on file and this creates lots of
+        # noise in the tests.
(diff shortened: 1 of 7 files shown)


choldgraf:
This one is a minor bugfix, so will self-merge if the tests and UX works as expected.

choldgraf:
Hmmm good point - it is a bit tricky to get just right, because if the top padding is too small then we will accidentally highlight the wrong section and the original problem will come back. I reduced the padding at the top in the latest commit to try to strike a balance. WDYT?

mathbunnyru:
I think it looks better now, but I still see the text from the previous part. I guess it's impossible to have both things.
Screenshot 2023-01-05 at 19 16 40

choldgraf:
OK how's this? It means that if we have two headers directly after one another, we might accidentally highlight the second, but maybe that's not such a big deal and also not common.

mathbunnyru:
Now it looks great to me 👍

PR #680, opened 2023-02-15
DONT MERGE: testing build package

AakashGfude:


choldgraf:
Updated from `master` here @AakashGfude - so you should be able to test that this now works

.github/workflows/publish.yml
```diff
@@ -23,11 +24,12 @@ jobs:
 
       - name: Build package
         run: |
+          sudo apt-get install gettext
```
choldgraf, line 27:
```suggestion
          # Required for translation infrastructure
          sudo apt-get install gettext
```

AakashGfude:
@choldgraf looks like it's working. I will close this PR for referencing and open a new one with just the required changes.

Issue #682, opened 2023-02-16
Renaming the in-page Table of Contents no longer works (0.4.0rc1)


jonascj:
### Describe the bug

Renaming the in-page Table of Contents as described here https://sphinx-book-theme.readthedocs.io/en/stable/customize/sidebar-secondary.html#rename-the-in-page-table-of-contents no longer works in 0.4.0rc1.

![Screenshot from 2023-02-16 13-15-49](https://user-images.githubusercontent.com/3301843/219362345-32ad9a80-c7ba-4f1b-8055-1a1c8521715a.png)

The HTML output apparently contains two versions of the in-page table of contents, one for printing and one for screens. The ``toc_title`` string is only updated in the one for printing, not the one for screens (see below excerpts below, one containing the string ``MY_TOC_TITLE`` the other ``On this page``).

**toc for print**
 ``` html
<div id="jb-print-docs-body" class="onlyprint">
    <h1>Installation og opsætning</h1>
    <!-- Table of contents -->
    <div id="print-main-content">
        <div id="jb-print-toc">
            
            <div>
                <h2> MY_TOC_TITLE </h2>
            </div>
            <nav aria-label="Page">
                <ul class="visible nav section-nav flex-column">
 <li class="toc-h2 nav-item toc-entry">
  <a class="reference internal nav-link" href="#installation-of-git">
   2.1. Installation of git
  </a>
  <ul class="visible nav section-nav flex-column">
   <li class="toc-h3 nav-item toc-entry">
    <a class="reference internal nav-link" href=#windows">
     2.1.1. Windows
```

**toc for screens**
``` html     
<div class="bd-sidebar-secondary bd-toc">
              
<div class="toc-item">
  
<div class="tocsection onthispage">
    <i class="fa-solid fa-list"></i> On this page
</div>
<nav id="bd-toc-nav" class="page-toc">
    <ul class="visible nav section-nav flex-column">
 <li class="toc-h2 nav-item toc-entry">
  <a class="reference internal nav-link" href="#installation-of-git">
   2.1. Installation of git
  </a>
  <ul class="visible nav section-nav flex-column">
   <li class="toc-h3 nav-item toc-entry">
    <a class="reference internal nav-link" href="#windows">
     2.1.1. Windows     
     
```

With version 0.4.0rc1 the default string ``CONTENTS`` is no longer shown on screens, it shows ``ON THIS PAGE`` instead. Presumably due to the same issue (only the toc for printing has its title string updated).

[...]

PR #683, opened 2023-02-16
FIX: Extra bugs after the theme switch

choldgraf:
Fixing a few extra bugs that were uncovered after updating to the latest pydata theme. This is mostly minimal bugfixes and style improvements. I'll plan to self-merge this once I've taken a few more passes, unless there is something complex in the implementation.

- closes https://github.com/executablebooks/sphinx-book-theme/issues/682


choldgraf:
I'm gonna merge this one so that we can make another pre-release, since we'll want changes to be tested out.

src/sphinx_book_theme/__init__.py
```diff
@@ -60,7 +60,7 @@ def add_metadata_to_page(app, pagename, templatename, context, doctree):
     context["translate"] = translation
 
     # If search text hasn't been manually specified, use a shorter one here
-    theme_options = app.config.html_theme_options or {}
+    theme_options = app.builder.theme_options or {}
```
mmcky, line 63: @choldgraf I think this has caused issues for the `pdflatex` builder (and will for other builders as well) as many builders won't have a `theme_options` object

Issue #584, opened 2022-07-10
Language support is broken - .mo files not bundled with this theme's wheel


whyjz:
### Describe the bug

**context and bug**
The language support seems not working (or partially working) now when used together with Jupyter Book. No matter what language is set in `_config.yml`, for example:
```
sphinx:
  config:
    language: fr
```
There will still be many words remaining untranslated (i.e. in English). Based on the following example, it seems that it used to work but not anymore.

### Likely diagnosis of problem

There are potentially two problems:

1. `.mo` files are not bundled with this theme. For example, [in this pre-release](https://pypi.org/project/sphinx-book-theme/0.4.0rc1) there are generated language files in `theme/sphinx_book_theme/static/locales/` but they are only `.po` and not `.mo`.

   Here's where the compilation happens:

   https://github.com/executablebooks/sphinx-book-theme/blob/cda3bb5753669c095903986cc0a649ff400fcef1/src/sphinx_book_theme/_compile_translations.py#L62-L73

   It isn't clear why this is happening - the only thing I can think of is that the `glob` isn't picking up the `.po` files at the end only in the build job for some reason? They are there when I build it locally.
2. Some theme wording has likely changed. For example, `Toggle primary sidebar` is (I think) a new phrase that is not in our [inventory of language translation phrases](https://github.com/executablebooks/sphinx-book-theme/tree/master/src/sphinx_book_theme/assets/translations). We should find the new phrases that aren't there, and add them via the instructions in that readme.

### Reproduce the bug

I forked a [Jupyter Book written in French](https://mi-gt-donnees.pages.math.unistra.fr/guide/00-introduction.html) (associated [Github repo](https://github.com/miti-gt-donnees/guide)) and pushed the book to my [Github pages](https://whyjz.github.io/miti-gt-donnees-guide/00-introduction.html) (associated [Github repo](https://github.com/whyjz/miti-gt-donnees-guide)) using exactly the same Github Action.

Screenshot from the original Book pages (last push was on Dec 8, 2021)
![1](https://user-images.githubusercontent.com/19339926/178138050-e87d2ce2-7fba-4020-96bd-607fe87ac9de.png)

Screenshot from the forked Book pages (last push was on Jul 10, 2022)
![2](https://user-images.githubusercontent.com/19339926/178138093-ce905168-bfd1-4343-a1ec-a8bf480f9e9e.png)

[...]

PR #684, opened 2023-02-17
FIX: Translation support and add new translations

AakashGfude:
- [x] fix `_compile_translations.py` , so that .mo files are present in build folder
- [ ] add missing translations

fixes https://github.com/executablebooks/sphinx-book-theme/issues/584

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,6 +23,7 @@ jobs:
 
       - name: Build package
         run: |
+          sudo apt-get install gettext
           python -m pip install -U pip build
           python -m build
 


AakashGfude:
@choldgraf it seems I can only translate to 9 languages from smodin. Do you have a premium account? As it seems other words have translations in more languages.

choldgraf:
Let's just go with the 9 languages - this is just a start anyway, I think that it will likely require manual updates for many languages anyway since automatic translating is likely to so things slightly incorrectly

AakashGfude:
ooh okay cool. I reckon we need to find words manually of what we want to update. Some words already found are:- 
`Last updated on` and `By` as mentioned in the issue https://github.com/executablebooks/sphinx-book-theme/issues/584 .  So, I will just go with that for now.

choldgraf:
Yeah that makes sense - we can do the best we can now, and if we miss something we can always improve it in a future PR !

choldgraf:
By the way, do you know how to handle translation strings when there is a word that isn't supposed to be translated? Eg "edit on GitHub" - we want "GitHub" to be part of the phrase (so that for example German speakers see it at the beginning of the phrase instead of the end), but it isn't translatable. I think there are guidelines for how to handle this but I don't know what they are

AakashGfude:
By the way, `Last updated on` and `By` are already here https://github.com/executablebooks/sphinx-book-theme/tree/master/src/sphinx_book_theme/assets/translations/jsons . So, I guess we don't need to add any translations at the moment?

choldgraf:
Ah nice - so we don't need to change anything for now? If that's the case I think we should just merge this and cut a release candidate

AakashGfude:
> Ah nice - so we don't need to change anything for now?
Seems like it. But I don't know why it was not working in the first place. Because of lack of `.mo` files?

choldgraf:
Yes without those files present six
Sphinx cannot generate any translations

PR #692, opened 2023-02-24
Add the possibility to disable launch buttons from Jupyter Book

paugier:
Fixes https://github.com/executablebooks/jupyter-book/issues/1463

### DIFF
--- a/src/sphinx_book_theme/header_buttons/launch.py
+++ b/src/sphinx_book_theme/header_buttons/launch.py
@@ -41,7 +41,11 @@ def add_launch_buttons(
     # If so, insert the URLs depending on the configuration
     config_theme = app.config["html_theme_options"]
     launch_buttons = config_theme.get("launch_buttons", {})
-    if not launch_buttons or not _is_notebook(app, pagename):
+    if (
+        not launch_buttons
+        or not _is_notebook(app, pagename)
+        or launch_buttons.get("disable", False)
+    ):
         return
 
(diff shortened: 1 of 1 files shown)


choldgraf:
This seems like a reasonable fix to me, but I wanted to throw out one other idea because it might be cleaner:

The goal is for launch buttons only to show up if a user has configured them to do so. I think the real problem is these things:

- Jupyter book sets "mybinder.org" as the default binder instance.
- The check for whether the buttons should exist only checks for the presence of _any_ launch button config, as opposed to specifically checking whether a launch button service has been requested (eg binder, colab, etc)

So a way to fix this without changing config values would be to make launch buttons only show if at least one service is explicitly specified, rather than defaulting to mybinder.org. it would mean that users explicitly have to put mybinder.org, but maybe that's ok?

What do you think?

paugier:
I don't have strong opinion on what is better ("manually disable" or "manually enable").

An advantage of the solution "manually disable" is that it is not a breaking change. Most users won't be affected by this change.

choldgraf:
Now that I think about it more, I think we want the default in JB to be "no launch buttons unless you explicitly define a service". That's because otherwise there is no guarantee that the repository is actually buildable on mybinder.org . So we don't want to show a button to use binder if we don't know binder is even an option

choldgraf:
closing this since it was superceded by #693 and the change to jupyter book

PR #691, opened 2023-02-23
NEH: Icon links, latest PST, more source providers, and bug fixes

choldgraf:
This brings in the latest pydata theme, and along the way includes support for their `icon-links` buttons in our primary sidebar. We just inherit the component template and have some lightweight styling to make it look in-line with our documentation style.

- Brings in some more fixes from feedback in https://github.com/executablebooks/sphinx-book-theme/issues/686
- closes #285 
- Fixes the bug about `builder.theme_options` that @mmcky reported
- closes https://github.com/executablebooks/sphinx-book-theme/issues/694
- closes https://github.com/executablebooks/sphinx-book-theme/issues/606

Also adds support for more source button providers via the pydata theme:

- closes https://github.com/executablebooks/sphinx-book-theme/issues/418
- closes https://github.com/executablebooks/sphinx-book-theme/issues/114
- closes https://github.com/executablebooks/sphinx-book-theme/issues/564

PR #693, opened 2023-02-25
Better early return for add_launch_buttons if no launch buttons

paugier:
Fixes https://github.com/executablebooks/jupyter-book/issues/1463

See https://github.com/executablebooks/jupyter-book/issues/1463#issuecomment-1445131176

Is it enough?

### DIFF
--- a/src/sphinx_book_theme/header_buttons/launch.py
+++ b/src/sphinx_book_theme/header_buttons/launch.py
@@ -41,7 +41,14 @@ def add_launch_buttons(
     # If so, insert the URLs depending on the configuration
     config_theme = app.config["html_theme_options"]
     launch_buttons = config_theme.get("launch_buttons", {})
-    if not launch_buttons or not _is_notebook(app, pagename):
+    if (
+        not launch_buttons
+        or not _is_notebook(app, pagename)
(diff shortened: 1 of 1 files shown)


choldgraf:
Nice - this looks correct to me. Could you just add a test to make sure that if none if them are there, then there is no launch buttons dropdo container on the page?

src/sphinx_book_theme/header_buttons/launch.py
```diff
@@ -41,7 +41,14 @@ def add_launch_buttons(
     # If so, insert the URLs depending on the configuration
     config_theme = app.config["html_theme_options"]
     launch_buttons = config_theme.get("launch_buttons", {})
-    if not launch_buttons or not _is_notebook(app, pagename):
+    if (
+        not launch_buttons
+        or not _is_notebook(app, pagename)
+        or not any(
+            launch_buttons[key]
```
choldgraf, line 48: Shouldnt we use `launch_buttons.get(key)` in case one is not present so we don't raise a key error?

choldgraf:
actually, no worries I will add a test in #691 - many thanks for this improvement :-)

PR #701, opened 2023-03-01
RLS 1.0.0: (repeat) Fix a link and improve test reliability

choldgraf:
We got bitten by a codecov rate limit bug (cc @AakashGfude you mentioned this in slack too) and then got bitten by a linkcheck that broke only _after_ the release had been made. I don't think we want to block release actions on broken links, so this just runs the codecov + the linkcheck jobs on the `pull_request` event, which means it won't trigger on the `release` event.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,9 +58,9 @@ jobs:
       run: >
         pytest --durations=10 --cov=sphinx_book_theme --cov-report=xml --cov-report=term-missing
 
-    # Only check for broken links on pull requests so that we don't trigger rate limit blocks
+    # Only upload to codecov on pull requests so that we don't trigger rate limit blocks
     - name: Upload to Codecov
-      if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.7 && github.repository == 'executablebooks/sphinx-book-theme && github.event_name == 'pull_request''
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -58,8 +58,9 @@ jobs:
       run: >
         pytest --durations=10 --cov=sphinx_book_theme --cov-report=xml --cov-report=term-missing
 
+    # Only check for broken links on pull requests so that we don't trigger rate limit blocks
```
choldgraf, line 61:
```suggestion
    # Only upload to codecov on pull requests so that we don't trigger rate limit blocks
```

.github/workflows/tests.yml
```diff
@@ -58,8 +58,9 @@ jobs:
...
     - name: Upload to Codecov
-      if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.7 && github.repository == 'executablebooks/sphinx-book-theme'
+      if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.7 && github.repository == 'executablebooks/sphinx-book-theme && github.event_name == 'pull_request''
```
choldgraf, line 63:
```suggestion
      if: matrix.os == 'ubuntu-latest' && matrix.python-version == 3.7 && github.repository == 'executablebooks/sphinx-book-theme' && github.event_name == 'pull_request'
```

choldgraf:
I'll wait for the linkcheck to complete on this PR, and also to confirm that codecov also runs, and will merge if so. Then we can double check that the jobs _don't_ run on `main`. If so I'll cut a new release

PR #741, opened 2023-06-03
update python requirements according to NEP 29

akhmerov:


choldgraf:
Thanks for this! Will just merge this one in and fix the tests in #766 since I think the python versions are triggering unexpected behavior

Issue #541, opened 2022-03-28
Side bar toggle not being inhereted?


da5nsy:
### Describe the bug

**context**
https://deploy-preview-2309--the-turing-way.netlify.app (as part of https://github.com/alan-turing-institute/the-turing-way/pull/2309)

**expectation**
I had assumed that the sidebar toggle would be inherited when moving from page-to-page within a site, but this does not seem to be happening?

### Reproduce the bug

- scroll to the bottom of any page
- toggle off the left sidebar
- click the 'next' button in the bottom right of the page
- watch as the sidebar returns on the next page

### List your environment

https://deploy-preview-2309--the-turing-way.netlify.app (as part of https://github.com/alan-turing-institute/the-turing-way/pull/2309)

choldgraf:
Just realized I never responded to this - thanks @da5nsy for reporting. I'm flagging it as an `enhancement` rather than a bug since I think it would be extending the functionality of this theme.

I think the way to do this would be to use browser storage that records the state of the sidebar and persists it across page visits. For example, in the pydata theme:

Here we set the position of the sidebar scroll:

https://github.com/pydata/pydata-sphinx-theme/blob/eff4dcd45a8403a57ffc116207062786abd46e52/src/pydata_sphinx_theme/assets/scripts/pydata-sphinx-theme.js#L161-L165

Here we grab this value and use it to scroll the sidebar to the appropriate location:

https://github.com/pydata/pydata-sphinx-theme/blob/eff4dcd45a8403a57ffc116207062786abd46e52/src/pydata_sphinx_theme/assets/scripts/pydata-sphinx-theme.js#L133-L136

ogrisel:
As an alternative to saving and re-setting the collapse state and scroll position, I think it would be nice to have a way to:

- expand only the ancestors of the current / active page;
- automatically scroll the left navbar so that the inner active header is at the top 25% of the page.

This way, when you click on a link to a specific jupyter-book page from an external website, the state of the left navbar would be intuitive.

Furthermore, when clicking "next" several times at the bottom of the main content area of jupyter book pages, the left side bar would progressively be updated to track the progress in the overall table of contents.

ogrisel:
Actually, re-reading the code of the `scrollToActive` function in `pydata-sphinx-theme.js` it already does what I want.

[...]

PR #754, opened 2023-08-30
FIX set bd-docs-nav as a class attribute instead of an id to restore compat with pydata-sphinx-theme JS

ogrisel:
This is a partial fix for #541.

It makes the `scrollToActive` function from `pydata-sphinx-theme.js` work as expected on jupyter-book rendered HTML.

However, this function does not save the collapsed state of other sections which make the "scroll to saved location" a bit weird/broken at times.

Still, it's already much better than the current state.


choldgraf (approved):
Ah yes - this looks good, thanks for noticing and making the PR!

PR #771, opened 2023-10-23
Make article width take up whole page in absence of secondary sidebar

### LINKED ISSUE #736: No right sidebar page width
### Describe the bug

When there is no header for a right sidebar, the width of the main content collapses

https://docs.scvi-tools.org/en/latest/api/index.html

Screenshot 2023-05-25 at 9 29 13 AM

### Reproduce the bug

Have a simple page with only a top-level heading.

### List your environment

_No response_

flying-sheep:
Fixes #736

<table>
<tr>
<tr>
<td>before:
<td>image
<tr>
<tr>
<td>after:
<td>image
</table>

### DIFF
--- a/src/sphinx_book_theme/assets/styles/sections/_article-container.scss
+++ b/src/sphinx_book_theme/assets/styles/sections/_article-container.scss
@@ -1,10 +1,13 @@
 .bd-main .bd-content .bd-article-container {
   // Re-adjust padding defaults to be flush on the top and right
   padding: 0rem;
-
   // Unset overflow x so that sticky: top works for the article header
   overflow-x: unset;
-  min-width: 0; // prevent from overflowing the flex container
+  // prevent from overflowing the flex container
+  min-width: 0;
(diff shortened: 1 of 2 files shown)


choldgraf (approved):
Thanks for this improvement - could you please explain the rationale behind the fix? I believe that it works, but I don't understand why setting a _max width_ will improve the case where the content was _too narrow_. How does a maximum width cause a shrunken container to grow?! Is this some kind of CSS quirk?

flying-sheep:
Before my change, `_article.scss` contained the `flex-grow: 0` hack. The hack was there so the absence of a sidebar doesn’t result in a super wide content area, but it had the side effect of making the content area collapse to a super narrow width in that case.

Removing that hack makes it fill more than the absolute minimum horizontal space. I then added `max-width` for making it not take any more space than it should and `justify-content: left` for not moving to the right. Those three changes replace the hack with a more robust solution.

choldgraf:
this is really helpful context, thanks for sharing!

Issue #765, opened 2023-10-05
PyData theme CSS resulted in unexpected changes to this theme's style


Gouvernathor:
Recent changes to the style have made the dark theme's background bluish instead of dark grey, and the hovered links purple instead of blue.
The preceding style was much better so I believe that change should be reverted, or at least an option provided to opt-out and go back to the former colors.

Former :
![image](https://github.com/executablebooks/sphinx-book-theme/assets/44340603/ea7f30b9-ff23-4c1f-b5d3-9b4dca0ac15a)
New :
![image](https://github.com/executablebooks/sphinx-book-theme/assets/44340603/747975de-fab7-4cdb-a2e2-8fb4265423db)

Gouvernathor:
I also wouldn't mind disabling the "Back to top" button spawning in the bottom half of the screen, at lease when you're not on a phone. It's not strictly related to the theme color issue, but I don't want to spam the tracker.

choldgraf:
A lot of these were inherited from the PyData Sphinx Theme, which made these updates to improve the accessibility of the theme in general. I don't think we should revert changes that we know improve accessibility (e.g. color contrast) just because we think they don't look as good, but if we can identify improvements that are equally as accessible, then I'm fine implementing them here.

Gouvernathor:
With the font color remaining stark white, since turning the background from pitch black to dark blue _decreases_ the contrast between them, I don't understand how it _improves_ accessibility,
I can understand accessibility reasons for having the hovered links purple instead of brighter blue, it feels weird since it's a deviation from the standard but if the accessibiliity reasons are solid I guess you have to start somewhere.

For the "back to top" button, I doubt that's related (but I don't think you were saying that).

choldgraf:
I just added some changes to address some of these in #775 - can you take a look there? I think it'll be more productive to make some minor adjustments to this theme rather than to re-litigate the long accessibility discussions that happened already in the pydata theme

Issue #761, opened 2023-09-17
Not all CSS is applied with latest version of pydata-sphinx-theme


lucaspcamargo:
### Describe the bug

When I regenerated my website a bit earlier today...
I expected it to look the same as always.

However, some CSS fails to apply. I see a blueish background color, purple link underlines from pydata-sphinx-theme, and larger icons.

Maybe something changed with the latest version of pydata-sphinx-theme which is breaking the latest version of this theme.

### Reproduce the bug

I believe that regenerating a page with the latest version of pydata-sphinx-theme should be enough to reproduce the issue.

When I add `pydata-sphinx-theme==0.13.3` to my requirements file, as a current workaround, the issue vanishes.

### List your environment

Here is my venv:

```
ablog==0.11.5
accessible-pygments==0.0.4
alabaster==0.7.13
Babel==2.12.1
beautifulsoup4==4.12.2
certifi==2023.7.22
charset-normalizer==3.2.0
docutils==0.19
feedgen==0.9.0
idna==3.4
imagesize==1.4.1
insegel==1.3.1
invoke==2.2.0
Jinja2==3.1.2
livereload==2.6.3
lxml==4.9.3
markdown-it-py==3.0.0
MarkupSafe==2.1.3
mdit-py-plugins==0.4.0
mdurl==0.1.2
myst-parser==2.0.0
packaging==23.1
pydata-sphinx-theme==0.14.0
Pygments==2.16.1
python-dateutil==2.8.2
PyYAML==6.0.1
requests==2.31.0
six==1.16.0
snowballstemmer==2.2.0
soupsieve==2.5
Sphinx==6.2.1
sphinx-book-theme==1.0.1
sphinx-favicon==1.0.1
sphinx-notfound-page==1.0.0
sphinx-sitemap==2.5.1
sphinxcontrib-applehelp==1.0.7
sphinxcontrib-devhelp==1.0.5
sphinxcontrib-htmlhelp==2.0.4
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==1.0.6
sphinxcontrib-serializinghtml==1.1.9
tornado==6.3.3
typing_extensions==4.7.1
urllib3==2.0.4
watchdog==3.0.0
```

pietronvll:
Same issue here for a documentation hosted on Read The Docs. Adding `pydata-sphinx-theme==0.13.3` _before_ `sphinx-book-theme` in the requirements file solves the problem.

Issue #746, opened 2023-06-30
Move the search icon back to the sidebar to make it easier to find and use


choldgraf:
In 1.0, we adopted the PyData Sphinx Theme "in-page search icon" for our search:

<img width="584" alt="image" src="https://github.com/executablebooks/sphinx-book-theme/assets/1839645/fac2bc1c-3f06-4d47-a748-0e5c622a1d73">

However a number of folks have noted that this is a difficult place to notice, and it degrades the UX of finding and using search. Instead, we should default our search bar to be in the sidebar where it often is for these kinds of single sidebar themes.

This PR in the PyData theme makes this pretty easy to do here:

- https://github.com/pydata/pydata-sphinx-theme/pull/1233

PR #775, opened 2023-11-08
ENH: Add search field back to sidebar

choldgraf:
This makes minor CSS adjustments to fix some changes made in the latest pydata theme update.

Brief summary:

- Changes the default behavior of our search component. The **icon button** will show up on narrow screens, and the **search field button** will show up in the sidebar on wider screens.
- Re-applies the same primary/secondary colors to keep this consistent. It uses a slightly lighter color on dark mode to meet a11y contrast requirements (I used [this little tool for this](https://coolors.co/contrast-checker/176de8-ffffff)).
- De-saturates the background on dark mode (cc @Gouvernathor)
- Hides the "back to top" button on wide screens when the TOC of the page is present to avoid imposing it on the content.
- Standardizes the size of header buttons that use SVGs
- Fixes a regression in https://github.com/executablebooks/sphinx-book-theme/pull/771 for content width on narrow screens

Issues resolved:

- closes https://github.com/executablebooks/sphinx-book-theme/issues/746
- resolves https://github.com/executablebooks/sphinx-book-theme/issues/761
- resolves https://github.com/executablebooks/sphinx-book-theme/issues/765

I think this is ready to merge, the two failures aren't related to this PR...

- I have a PR to fix the docs audit failure: https://github.com/pydata/pydata-sphinx-theme/pull/1556
- And an issue related to the windows build failure: https://github.com/pradyunsg/sphinx-theme-builder/issues/46

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
     - id: flake8
 
   - repo: https://github.com/psf/black
-    rev: 23.9.1
+    rev: 23.10.1
     hooks:
     - id: black
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,15 +22,12 @@ repos:
     - id: end-of-file-fixer
     - id: trailing-whitespace
 
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.1.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.1.4
     hooks:
-    - id: flake8
-
-  - repo: https://github.com/psf/black
-    rev: 23.10.1
-    hooks:
-    - id: black
+    - id: ruff
+      args: ["--fix", "--show-fixes"]
+    - id: ruff-format
 
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v3.0.3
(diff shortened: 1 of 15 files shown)


choldgraf:
@Gouvernathor I've got some changes in this one that are relevant to you, if you want to take a look

Gouvernathor:
That sounds great, how can I test how it renders ?

choldgraf:
I think for now your best bet is to look at the rendered documentation preview from this PR.

https://sphinx-book-theme--775.org.readthedocs.build/en/775/

It will be way easier to just check that out and estimate how it would look rather than to build your docs locally with this PR.

Gouvernathor:
Thanks ! It's more than enough.
As for the background color, I would prefer the former color, `#121212`, which is [visually much darker](https://sphinx-book-theme.readthedocs.io/en/stable/) than [this new one](https://sphinx-book-theme--775.org.readthedocs.build/en/775/). Even darker would be acceptable too, for me, but a simple rollback to the previous color (which is still part of the version of book that's considered "stable") may be more acceptable for users at-large,
The "back to top" and research bar changes are great.

choldgraf:
Fair enough - I think keeping the old background color is less messy anyway. I'd rather not change something if we don't have a clear rationale for doing so. I've reverted that back in the last commit

choldgraf:
OK going to merge this one in as soon as the audit test is happy. I think it's a good enough improvement and is another step to unblocking folks with all the upstream updates.

PR #791, opened 2023-12-12
FIX: properly handle CSS objects in css_files

agoose77:
Sphinx 7.26 (approximately) deprecates implicit string operations on the contents of `context['css_files']`. This, in turn, raises warnings for users that enable strict warning checks. This was observed in `jupyter-book`'s test-suite.

This PR updates our handling to first assume objects within this list are `_CascadingStyleSheet` objects, and replaces the object back on the path.

### DIFF
--- a/src/sphinx_book_theme/__init__.py
+++ b/src/sphinx_book_theme/__init__.py
@@ -77,7 +77,7 @@ def _gen_hash(path: str) -> str:
     return hashlib.sha1(path.read_bytes()).hexdigest()
 
 
-def hash_assets_for_files(assets: list, theme_static: Path, context):
+def hash_assets_for_files(assets: list, theme_static: Path, context, app):
     """Generate a hash for assets, and append to its entry in context.
 
     assets: a list of assets to hash, each path should be relative to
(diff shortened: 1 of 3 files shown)


src/sphinx_book_theme/__init__.py
```diff
@@ -88,22 +88,36 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
...
             if not asset_source_path.exists():
                 SPHINX_LOGGER.warning(
                     f"Asset {asset_source_path} does not exist, not linking."
                 )
             # Find this asset in context, and update it to include the digest
-            if asset_sphinx_link in context[asset_type]:
-                hash = _gen_hash(asset_source_path)
-                ix = context[asset_type].index(asset_sphinx_link)
-                context[asset_type][ix] = asset_sphinx_link + "?digest=" + hash
+            for i, other_asset in enumerate(context[asset_type]):
```
choldgraf, line 103: nitpick but I'm a fan of `ii` instead of `i` so that it's easier to ctrl+f if need be 🤷 . Disregard this if you wish!

choldgraf (approved):
This looks good to me - a minor nitpick but LGTM regardless.

Should double-check that the "unexpected warnings" test failure is real or not though

src/sphinx_book_theme/__init__.py
agoose77: Fine, no strong feelings!

src/sphinx_book_theme/__init__.py
```diff
@@ -88,22 +88,36 @@ def hash_assets_for_files(assets: list, theme_static: Path, context):
...
+                # Take priority from existing asset or use default priority (500)
+                priority = getattr(other_asset, "priority", 500)
+                # Remove existing asset
+                del context[asset_type][i]
+                # Add new asset
+                app.add_css_file(
+                    asset_sphinx_link,
+                    digest=_gen_hash(asset_source_path),
+                    priority=priority,
+                )
```
agoose77, line 120:
```suggestion
            for ii, other_asset in enumerate(context[asset_type]):
                # TODO: eventually the contents of context['css_files'] etc should probably
                #       only be _CascadingStyleSheet etc. For now, assume mixed with strings.
                if (
                    getattr(other_asset, "filename", str(other_asset))
                    != asset_sphinx_link
                ):
                    continue
                # Take priority from existing asset or use default priority (500)
                priority = getattr(other_asset, "priority", 500)
                # Remove existing asset
                del context[asset_type][ii]
                # Add new asset
                app.add_css_file(
                    asset_sphinx_link,
                    digest=_gen_hash(asset_source_path),
                    priority=priority,
                )
```

src/sphinx_book_theme/__init__.py
```diff
@@ -117,7 +131,7 @@ def hash_html_assets(app, pagename, templatename, context, doctree):
     # run but the book theme CSS file won't be linked in Sphinx.
     if app.config.html_theme == "sphinx_book_theme":
         assets.append("styles/sphinx-book-theme.css")
-    hash_assets_for_files(assets, get_html_theme_path() / "static", context)
+    hash_assets_for_files(assets, get_html_theme_path() / "static", context, app)
```
agoose77, line 134: This function is public, but I can't imagine anyone using it. Are you happy with the breaking change?

PR #800, opened 2023-12-20
DOCS: require compatible docutils

### LINKED ISSUE #799: Broken page on the website with the documentation
### Describe the bug

https://sphinx-book-theme.readthedocs.io/en/latest/reference/extensions.html

image

Google Chrome

### Reproduce the bug

Open a page in a browser
https://sphinx-book-theme.readthedocs.io/en/latest/reference/extensions.html

### List your environment

_No response_

agoose77:
The widespread problem with docutils 0.17 and 0.18 also affects our docs generation. As our optional `docs` extra is intended for internal use only, we can describe the compatibility safely in `pyproject.toml` rather than worrying about backsolving.

Fixes #799

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ doc = [
     "sphinxcontrib-bibtex",
     "sphinxcontrib-youtube",
     "sphinxext-opengraph",
+    "docutils!=0.17.*,!=0.18.*"
 ]
 test = [
     "beautifulsoup4",


choldgraf:
Looks like we're using a deprecated function (`add_javascript` I believe is now `add_js_file`). 

Also that website preview is pretty glorious haha

agoose77:
The issue here is the widespread docutils div bug, and we're seeing it because sphinx-tabs doesn't yet support sphinx 7 (thus everything is downgraded). As a result, the fix is really in Sphinx-tabs (I have an open pr there)!

choldgraf:
Ah yes the docutils gift that keeps on giving

PR #806, opened 2024-01-05
Edited russian translate "By"

askras:


agoose77:
@askras does this translation make sense if you remove the trailing colon?

askras:
> @askras does this translation make sense if you remove the trailing colon?

Good afternoon!

If you remove the colon, the meaning remains. But with a colon it’s more correct

askras:
Perhaps it is even better to use an empty string as a translation ". Then everyone will be able to use their own format of the author's description.

PR #813, opened 2024-02-07
refactor: handle options more pythonically

agoose77:


choldgraf (approved):
@agoose77 in the future feel free to just merge these on your own if you don't think it's a controversial or complex change

Issue #722, opened 2023-04-14
Typo in tooltip text of launchers


rhugonnet:
Hi all,

First, thanks for the amazing work on this theme! :slightly_smiling_face: 

Just noticed a very minor issue while hovering over the Binder launcher, there is a space missing in the tooltip, it shows as: `Launch onBinder`

Looks like this needs to be corrected for all launcher types: 
https://github.com/executablebooks/sphinx-book-theme/blob/c2664ddfbe2c3fcfaa819a57835f00ca58346cb9/src/sphinx_book_theme/header_buttons/launch.py#L136
https://github.com/executablebooks/sphinx-book-theme/blob/c2664ddfbe2c3fcfaa819a57835f00ca58346cb9/src/sphinx_book_theme/header_buttons/launch.py#L154
https://github.com/executablebooks/sphinx-book-theme/blob/c2664ddfbe2c3fcfaa819a57835f00ca58346cb9/src/sphinx_book_theme/header_buttons/launch.py#L169
https://github.com/executablebooks/sphinx-book-theme/blob/c2664ddfbe2c3fcfaa819a57835f00ca58346cb9/src/sphinx_book_theme/header_buttons/launch.py#L185

Or maybe I don't fully get the behaviour of `sphinx.locale.get_translation`?
Happy to open a PR if that saves time.

**On another topic**: Any update/plan to work on https://github.com/executablebooks/sphinx-book-theme/issues/282? I'm also very interested in that feature, could potentially contribute if nothing is ongoing already.

PR #831, opened 2024-04-08
[FIX] fixes to the dropdown buttons

afeld:
This pull request includes a handful of improvements around the drop-down/launch buttons:

- Improves the interactivity of the drop-down buttons: Version 1.0.0 introduced a gap between the hover area and the drop-down. When the mouse would move from the former to the latter (especially if slowly), the drop-down would disappear. You can see this behavior on [the live site](https://sphinx-book-theme.readthedocs.io/en/stable/content/notebooks.html). cc @choldgraf since you did work in this area.
- Adds alt text to the launch button logos
- Adds a space between "Launch on" and the tool name - fixes https://github.com/executablebooks/sphinx-book-theme/issues/722 cc @rhugonnet 

Screenshot showing "Launch on Colab"

### DIFF
--- a/src/sphinx_book_theme/assets/styles/sections/_header-article.scss
+++ b/src/sphinx_book_theme/assets/styles/sections/_header-article.scss
@@ -91,7 +91,8 @@ label.sidebar-toggle.secondary-toggle {
     font-size: 1.3rem;
     color: var(--pst-color-text-muted);
     border: none;
-    padding: 0 0.5rem;
+    padding-left: 0.5rem;
+    padding-right: 0.5rem;
 
     // Make sure anything inside is aligned vertically
     display: flex;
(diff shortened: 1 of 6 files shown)


agoose77:
@choldgraf LGTM, but do cast your eye over the SBT dropdowns for "Launch on"; I didn't author this originally, so I'm not familiar with whether this is a regression or a long-standing bug.

afeld:
Looks like it was introduced with the translation in https://github.com/executablebooks/sphinx-book-theme/commit/1656b772f78153e18bebde637c27482e30d4365c#diff-c18ac40355c51c6c6566efc355dfa5e7f002859040252df71ebc2d3a01da96bcL115.

choldgraf (approved):
This all looks great to me - I'm not sure what bug @agoose77 is pointing out, but I agree I don't think it's related to this PR.

I'll merge this one in, and @agoose77 if there's something you think is wrong, could you open an issue that describes it so we can spot-check in a follow up?

PR #844, opened 2024-06-12
really drop python<=3.8 support

kloczek:
Filter all code over `pyupgrade --py39-plus`.


agoose77:
This has been on my To Do list for a while! Thanks!

PR #867, opened 2024-11-19
Exclude command prompts when using copy button

stevepiercy:
See:
- https://sphinx-copybutton.readthedocs.io/en/latest/use.html#automatic-exclusion-of-prompts-from-the-copies
- https://github.com/pydata/pydata-sphinx-theme/pull/2036

Alternatively, don't use prompts in shell commands.


agoose77:
Thanks for this @stevepiercy, a good suggestion!

PR #882, opened 2024-12-22
Make the generated gettext files reproducible

mitya57:
I am packaging sphinx-book-theme for Debian, and one of our goals is to make the built packages reproducible. See https://reproducible-builds.org/ for details.

In this case, the generator of `.po` files depended on file system order, which may vary between different machines and file system types.


choldgraf (approved):
this is reasonable - thanks

PR #883, opened 2024-12-22
Make the generated documentation more reproducible

mitya57:
I am packaging sphinx-book-theme for Debian, and one of our goals is to make the built packages reproducible. See https://reproducible-builds.org/ for details.

This pull request makes the following improvements:

- Use static ID for tables generated by pandas. This removes the following diff:
  ```diff
  -<table id="T_08154" style="font-size: 10px">
  +<table id="T_afb07" style="font-size: 10px">
  ```
  We have only one table on the page, so we can just hardcode `1`. Also, replace deprecated `applymap` with `map`.

- Use `plt.show()` as the final statement of matplotlib blocks, to prevent repr of AxesImage being added to HTML, which causes this diff:
  ```diff
  -<matplotlib.image.AxesImage at 0x7fffddde12b0>
  +<matplotlib.image.AxesImage at 0x7f38706eade0>
  ```

- Call `np.random.seed(0)` before using `np.random.randn()`.

After these fixes there are still problems coming from Folium and from equation with automatic label, but I have found no good solution, so I will probably just patch them out in Debian.

### DIFF
--- a/docs/content/notebooks.md
+++ b/docs/content/notebooks.md
@@ -171,6 +171,7 @@ See [the Pandas Styling docs](https://pandas.pydata.org/pandas-docs/stable/user_
 :tags: [hide-input]
 
 import pandas as pd
+from pandas.io.formats.style import Styler
 
 np.random.seed(24)
 df = pd.DataFrame({'A': np.linspace(1, 10, 10)})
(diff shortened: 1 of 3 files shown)


choldgraf (approved):
This all looks reasonable to me - thanks for these improvements to make it a bit more deterministic!

PR #916, opened 2025-07-11
Show source text on GitLab projects

tymcauley:
This aligns with the behavior for GitHub projects. Here's an example of a GitLab-hosted Markdown file showing source code with the `?plain=1` parameter: https://gitlab.com/gitlab-org/cli/-/blob/main/README.md?plain=1


choldgraf:
This looks good to me, thanks for standardizing behavior with the github workflow!

PR #927, opened 2026-03-02
MAINT: adding sphinx9 and python3.14 support and bumping minimum sphinx to 7 and python to 3.11

bsipocz:
This PR should get to the point of passing CI before merging; then we'll be ready for a release.

`ablog` pin is for https://github.com/sunpy/ablog/issues/313 which also means that it's the cleanest and easiest to also bump the minimum dependencies to sphinx and python to something less ancient.


bsipocz:
OK, so apparently I need to do a `sphinx-tabs` release first.

PR #950, opened 2026-04-09
Test author against default value

cedk:
The default value for author is 'Author name not set'.


bsipocz:
Could you please bring an example where this causes issues? I don't particularly see where this default is coming from.

cedk:
The default value comes from "Sphinx", see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-author
So without explicitly set, the template render "By Author name not set" as on https://docs.tryton.org/latest/

